PR #29, opened 2024-10-25
26 Variable importance: support 'decision_fn' 'transform' methods

jpaillard:
In addition to the minor modification allowing to use more methods I added a detailed example of variable inference on a classification problem. 
It highlights: 
 - The benefits of having model-agnostic methods: here linear SVC fails when using a polynomial kernel works
 - The benefits of using conditional importance: The features are correlated making PI identify spurious important features. 
 - That the current implementation supports `decision_function` to score importance

### DIFF
--- a/examples/plot_variable_importance_classif.py
+++ b/examples/plot_variable_importance_classif.py
@@ -235,7 +235,7 @@ box1 = ax.boxplot(
     importance_arr[:, 0, :],
     positions=np.arange(1, n_features + 1) - 0.25,
     widths=0.2,
-    label="Linear",
+    label="CPI Linear",
     vert=False,
 )
 for item in ["whiskers", "fliers", "medians", "caps", "boxes"]:
@@ -245,7 +245,7 @@ box1 = ax.boxplot(
     importance_arr[:, 1, :],
     positions=np.arange(1, n_features + 1),
     widths=0.2,
-    label="Poly",
+    label="CPI Poly",
     vert=False,
 )
 for item in ["whiskers", "fliers", "medians", "caps", "boxes"]:
@@ -255,7 +255,7 @@ box1 = ax.boxplot(
     importance_arr[:, 2, :],
     positions=np.arange(1, n_features + 1) + 0.25,
     widths=0.2,
-    label="Permutation",
+    label="PI Poly",
     vert=False,
 )
 for item in ["whiskers", "fliers", "medians", "caps", "boxes"]:
--- a/examples/plot_variable_importance_classif.py
+++ b/examples/plot_variable_importance_classif.py
@@ -9,8 +9,8 @@ illustrating the importance of model-agnostic variable importance methods, which
 opposed to linear models for instance, can capture non-linear relationships. 
 The features are generated from a multivariate normal distribution with a Toeplitz 
 correlation matrix. This second specificity of the problem is interesting to exemplify 
-the benefits of the conditional permutation importance (CPI) method :footcite:t:`Chamma_NeurIPS2023` 
-over the standard permutation importance (PI) method :footcite:t:`breimanRandomForests2001`.
+the benefits of the conditional permutation importance (CPI) method [:footcite:t:`Chamma_NeurIPS2023`] 
+over the standard permutation importance (PI) method [:footcite:t:`breimanRandomForests2001`].
 
 References
 ----------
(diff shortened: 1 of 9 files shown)


jpaillard:
The doc looks good to me. I think this is ready 
ping @dengemann

dengemann:
Looks very cool, exactly what I had in mind!

hidimstat/cpi.py
```diff
@@ -170,10 +173,7 @@ def _joblib_predict_one_group(imputation_model, X, j):
                 if isinstance(X, pd.DataFrame):
                     X_perm = pd.DataFrame(X_perm, columns=X.columns)
 
-                if self.score_proba:
-                    y_pred_perm = self.estimator.predict_proba(X_perm)
-                else:
-                    y_pred_perm = self.estimator.predict(X_perm)
+                y_pred_perm = getattr(self.estimator, self.method)(X_perm)
```
dengemann, line 176: probably this should be handled explicitly

```Python

y_pred_perm = getattr(self.estimator, self.method, None)(X_perm)
if y_pred_perm is None:
    raise ValueError(r"The predict method ({self.method}) you passed is not supported by the estimator")
```

hidimstat/cpi.py
```diff
@@ -217,10 +217,7 @@ def score(self, X, y):
 
         out_dict = dict()
 
-        if self.score_proba:
-            y_pred = self.estimator.predict_proba(X)
-        else:
-            y_pred = self.estimator.predict(X)
+        y_pred = getattr(self.estimator, self.method)(X)
```
dengemann, line 220: also here. perhaps factorize it into a small getter function

hidimstat/cpi.py
dengemann: ok. Then fine!

jpaillard:
ping @bthirion

bthirion (commented):
Sorry, I had done the review a few days ago but forgot to submit it ...

jpaillard:
I added the test for Errors. 
This is ready on my side

examples/plot_variable_importance_classif.py
```diff
@@ -0,0 +1,299 @@
...
+model_non_linear = RandomizedSearchCV(
+    SVC(
+        random_state=seed,
+        kernel="poly",
+        degree=2,
+        coef0=1,
+    ),
+    param_distributions={
+        "C": np.logspace(-3, 3, 10),
+        # "gamma": np.logspace(-3, 3, 10),
```
bthirion, line 149: remove commented lines

examples/plot_variable_importance_classif.py
```diff
@@ -0,0 +1,299 @@
...
+        "Age",
+        "X8",
+        "X9",
+        "X10",
+    ]
+)
+tck_ids = np.arange(0, n_features)
+axes[1].set_xticks(tck_ids, labels[tck_ids], rotation=45)
+axes[1].set_yticks(tck_ids, labels[tck_ids])
+
```
bthirion, line 114: What do the red/blue dots correspond to ? Can you add a legend ?

examples/plot_variable_importance_classif.py
```diff
@@ -0,0 +1,299 @@
...
+        diff = importance_arr[:, i, j]
+        pval_arr[j, i] = ttest_1samp(diff, 0)[1]
+
+#############################################################################
+# Visualize the variable importance
+# ------------------------------
+# Here we plot the variable importance and highlight the features that are considered
+# important, with a p-value lower than 0.05, using a diamond marker. We also highlight
+# the true important features, used to generate the target variable, with a star marker.
+# While the linear model captures the importance of the age, it fails to capture the the
```
bthirion, line 226:
```suggestion
# While the linear model captures the importance of the age, it fails to capture the
```

bthirion (commented):
Minor improvements needed, but this looks very good overall !

examples/plot_variable_importance_classif.py
jpaillard: I updated the legend, it corresponds to variables for which the p-value is < threshold (here I used 0.05)

jpaillard:
I addressed your comments @bthirion

hidimstat/cpi.py
```diff
@@ -42,7 +44,7 @@ def __init__(
         imputation_model,
         n_permutations: int = 50,
         loss: callable = root_mean_squared_error,
-        score_proba: bool = False,
+        method: str = "predict",
```
bthirion, line 47: You need to update the docstring above, as well as for LOCO and PI.

bthirion (commented):
Could you just update the docstrings ? Then we can merge.
Thx,

hidimstat/cpi.py
jpaillard: done

bthirion (approved):

bthirion:
Merging now, thx.

PR #110, opened 2025-01-02
fix dependencies

jpaillard:
Fixes the mismatch between dependencies described in the readme and the `pyproject.toml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ pip install -e .
 ```
 joblib
 numpy
+panda
 scipy
 scikit-learn
 torch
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ numpy
 scipy
```
lionelkusch, line 35:
```suggestion
panda
scipy
```
It also missing the panda distribution.

lionelkusch (changes requested):
After this suggestion, it is fine.

lionelkusch (approved):

PR #159, opened 2025-02-20
switch to 3-clause BSD license

jpaillard:


bthirion (approved):

LICENSE
```diff
@@ -1,19 +1,25 @@
-BSD 2-Clause License
+BSD 3-Clause License
 
-Copyright (c) 2024, Mind-Inria
+Copyright (c) 2024, The hidimstat developers.
```
lionelkusch, line 3: There is a change in the copyright holder. 
Do you think that it's better? @bthirion

bthirion: Yes, we want the project to be open to non-Inria people.

PR #184, opened 2025-03-13
Fix broken links in API doc

jpaillard:


lionelkusch:
This is the actual link:
https://github.com/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/hidimstat/clustered_inference.py#L171-L292
From my understanding, it should be like this:
https://github.com/mind-inria/hidimstat/blob/main/src/hidimstat/clustered_inference.py 

I am not sure that your fix is working. 
Can you add a print before to return of the link in order to check if the link are correct and there is not error before?

lionelkusch (approved):

PR #200, opened 2025-04-01
Register missing class attributes

jpaillard:


bthirion (commented):
Sounds right so far.

jpaillard:
That's all on my side. It fixes the issue.

src/hidimstat/conditional_permutation_importance.py
```diff
@@ -75,6 +75,9 @@ def __init__(
             "categorical": imputation_model_categorical,
         }
         self.categorical_max_cardinality = categorical_max_cardinality
+        self.imputation_model_categorical = imputation_model_categorical
```
lionelkusch, line 78: The imputation models are already saved in a dictionary (lines 73-76):
```
 self.imputation_model = {
            "continuous": imputation_model_continuous,
            "categorical": imputation_model_categorical,
        }
```
I don't see the point to register them another time because any modification of one variable won't modify the second one. 
I think the best is to remove the lines 73 to 76 and to modify the rest of the code.

src/hidimstat/permutation_importance.py
```diff
@@ -53,6 +53,7 @@ def __init__(
             n_permutations=n_permutations,
         )
         self.rng = check_random_state(random_state)
```
lionelkusch, line 55: If you register, `random_state`, you should remove `self.rgn` if it's not necessary, in other cases, you can replace it with `self._rng`

jpaillard: done

src/hidimstat/conditional_permutation_importance.py
jpaillard: For the above reasons, we need to have explicitly named attributes (e.g., self.imputation_model_continuous). So, I replaced the `imputation_model ` dict.

src/hidimstat/conditional_permutation_importance.py
```diff
@@ -67,14 +68,11 @@ def __init__(
...
         self._list_imputation_models = []
-
-        self.imputation_model = {
-            "continuous": imputation_model_continuous,
-            "categorical": imputation_model_categorical,
-        }
         self.categorical_max_cardinality = categorical_max_cardinality
+        self.imputation_model_categorical = imputation_model_categorical
+        self.imputation_model_continuous = imputation_model_continuous
+        self.random_state = check_random_state(random_state)
```
lionelkusch, line 75: Can you do the check in the fit method and not in the __init__ method?
This is mostly done like this in scikitlearn.

lionelkusch (changes requested):

src/hidimstat/conditional_permutation_importance.py
jpaillard: yes, done

bthirion (approved):
LGTM, thx.

lionelkusch (approved):
Good and thank you

PR #206, opened 2025-04-04
DOC fix docstring in deparsified_lasso_pvalue

glemaitre:
On the rendered documentation, it seems that the missing return line makes the numpydoc parser to fail to parse properly and render the documentation.


bthirion (commented):
LGTM. Are you done with this one ?
Best,

PR #210, opened 2025-04-04
continue circleci setup

jeromedockes:


glemaitre:
I think that we will need some additinal environment variables:

https://github.com/scikit-learn-contrib/imbalanced-learn/blob/8189ba7fc1d936de244e120389802654b87b9c29/.circleci/config.yml#L8-L12

jeromedockes:
now we have a built doc in the artifact: https://output.circle-artifacts.com/output/job/7913c62b-5f80-4002-b546-dc2f8d9d6a4c/artifacts/0/doc_conf/index.html


so we need 

- the artifact redirector
- the push to hidimstat.github.io

Issue #219, opened 2025-04-07
Limit examples based on complex simulations


jpaillard:
To follow up on our discussion before and during the sprint about examples and that we should rely as much as possible on already existing datasets, I suggest replacing or removing the example "Measuring variable importance in classification." Besides relying on a complicated simulation scenario, it is also quite complex. It shows 3 things: 
 1. How to use variable importance & controlled variable importance methods for classification 
 2. Problems arising from model misspecification when measuring variable importance and motivates model-agnostic approaches. 
 3. Pitfalls of permutation importance and how to address it will CPI. 

I think that 3) is now addressed in #207

bthirion:
Thx for opening this.
What is your suggestion for 1. and 2. ?

jpaillard:
For 2. I suggest an example on the circles dataset where we could show that an ANOVA f-test leads to no important p-values but that using LOCO with an SVM makes discoveries. We could also add the interpretation of LOCO as a generalized ANOVA as described in https://onlinelibrary.wiley.com/doi/full/10.1111/biom.13392

For 3. #218 and the example above would be two examples showing how to use VIM for classification, one with CPI, the other with LOCO.

bthirion:
Alternatively, the XOR dataset would be good to  study model misspecification (eg linear models).

I also think that having an example doing classification would be good. It is typically harder to do VI analysis with classification, as the decision  function is not so sensitive to small weights variations.

bthirion:
I don't think we should combine them.

jpaillard:
I suggest that when the following PRs introducing the examples are merged, we open a new one to remove the current ""Measuring variable importance in classification." which relies on complex simulation. 

- [ ] #223 
- [ ] #229  
- [ ] #218

PR #223, opened 2025-04-08
Example motivating model agnostic variable selection on circles dataset

jpaillard:
This example uses the circles dataset where classes are not linearly separable. It illustrates that conditional independence testing using d0CRT fails to identify important variables because it is based on linear models. Then, it shows that using LOCO, we can measure "controlled variable importance" and thus select variables if the predictive model is not misspecified. 

It is related to #219 (motivating model-agnostic approaches).

### DIFF
--- a/examples/plot_model_agnostic_importance.py
+++ b/examples/plot_model_agnostic_importance.py
@@ -1,5 +1,5 @@
 """
-Model Agnostic Variable Selection when Linear Models Fail
+Variable Selection Under Model Misspecification
 =============================================================
 
 In this example, we illustrate the limitations of variable selection methods based on


examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+###############################################################################
+# Compute variable selection using d0CRT and then LOCO
+# ---------------------------------------------------------------------------
+# We first compute the p-values using d0CRT which performs a conditional independence
+# test (:math:`H_0: X_j \perp\!\!\!\perp y | X_{-j}`) for each variable. However,
+# this test is based on a linear model (LogisticRegression) and fails to reject the null
+# in the presence of non-linear relationships. We then compute the p-values using LOCO
+# and show that it can accommodate both linear and non-linear models. When using a
```
bthirion, line 73:
```suggestion
# with a linear, and then a non-linear model. When using a
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
...
+_, pval_dcrt, _ = dcrt_pvalue(
+    selection_features=selection_features,
+    X_res=X_residual,
+    y_res=y_res,
+    sigma2=sigma2,
+    fdr=0.05,
+)
+# Compute p-values using LOCO
+cv = KFold(n_splits=5, shuffle=True, random_state=0)
+model_svc = SVC(kernel="rbf", random_state=0)
```
bthirion, line 89:
```suggestion
non-linear_model = SVC(kernel="rbf", random_state=0)
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
...
+model_linear = LogisticRegressionCV(Cs=np.logspace(-3, 3, 5))
```
bthirion, line 90:
```suggestion
linear_model = LogisticRegressionCV(Cs=np.logspace(-3, 3, 5))
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
...
+        estimator=model_linear_, loss=log_loss, method="predict_proba", n_jobs=2
+    )
+    vim_svc = LOCO(
+        estimator=model_svc_, loss=hinge_loss, method="decision_function", n_jobs=2
+    )
+    vim_linear.fit(X[train], y[train])
+    vim_svc.fit(X[train], y[train])
+
+    importances_linear.append(vim_linear.importance(X[test], y[test])["importance"])
+    importances_svc.append(vim_svc.importance(X[test], y[test])["importance"])
```
bthirion, line 110:
```suggestion
    importances_non-linear.append(vim_svc.importance(X[test], y[test])["importance"])
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+_, pval_linear = ttest_1samp(importances_linear, 0, axis=0, alternative="greater")
+_, pval_svc = ttest_1samp(importances_svc, 0, axis=0, alternative="greater")
```
bthirion, line 114:
```suggestion
_, pval_non-linear = ttest_1samp(importances_svc, 0, axis=0, alternative="greater")
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,146 @@
+"""
+Model Agnostic Variable Selection when Linear Models Fail
+=============================================================
+
+In this example, we illustrate the limitations of variable selection methods based on
```
lionelkusch, line 5: Can you add 1 line description of the example because this is used by sphinx gallery when the cursor is on the example?

lionelkusch (commented):
Very nice example.

examples/plot_model_agnostic_importance.py
jpaillard: done

examples/plot_model_agnostic_importance.py
jpaillard: done

examples/plot_model_agnostic_importance.py
jpaillard: It would currently show, 
> "In this example, we illustrate the limitations of variable selection methods based on
linear models using ..."
I think it conveys the main idea; the only missing part, "on the circles dataset," is captured by the thumbnail image.

examples/plot_model_agnostic_importance.py
jpaillard: done

examples/plot_model_agnostic_importance.py
lionelkusch: ok

bthirion:
The example looks good, nice ! 

My only request is that you chunk the code into smaller blocks, so that you can give more pedagogical explanations. Think of a user who discovers that altogether.

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,166 @@
...
+model used, :math:`\mu_{-j}` the sub-model where the j-th variable is removed, and
+:math:`X^{-j}` the data with the j-th variable removed, the loss difference can be
+expressed as:
+
+.. math::
+    \psi_{j} = \mathbb{V}(y) \left[ \left[ 1 - \\frac{\mathbb{E}[(y - \mu(X))^2]}{\mathbb{V}(y)} \\right] - \left[ 1 - \\frac{\mathbb{E}[(y - \mu_{-j}(X^{-j}))^2]}{\mathbb{V}(y)} \\right] \\right]
+
+where :math:`\psi_{j}` is the LOCO importance of the j-th variable.
+
+References
```
lionelkusch, line 25: Can you add the Reference section at the end of the example?

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,166 @@
...
+    palette="muted",
+)
+ax.legend(title="Class")
+ax.set_xlabel("X1")
+ax.set_ylabel("X2")
+plt.show()
+
+
+###############################################################################
+# Compute variable selection using d0CRT
```
lionelkusch, line 67:
```suggestion
# Compute p-values using d0CRT
```

examples/plot_model_agnostic_importance.py
```diff
@@ -0,0 +1,166 @@
...
+)
+ax.legend()
+plt.show()
+
+
+#################################################################################
+# As expected, when using linear models (d0CRT and LOCO-linear) that are misspecified,
+# the varibles are not selected. This highlights the benefit of using model-agnostic
+# methods such as LOCO, which allows for the use of models that are expressive enough
+# to capture the underlying data generating process.
```
lionelkusch, line 166: I don't like the expression `data generating process` because we want to explain the data and not the generation of data.
Can you change it, please?

jpaillard: This phrasing comes from [Verdinelli et al.](https://arxiv.org/abs/2303.05981)

> Population level feature importance is concerned with measuring how important a particular feature is in the underlying data generating process

I think that they also mean "explain the data".

examples/plot_model_agnostic_importance.py
jpaillard: done

examples/plot_model_agnostic_importance.py
lionelkusch: Yes, you are right. It's just that, in the context of experimental data, we often want to understand the underlying process which generates the data. However, when the data are synthetic, it can be a bit confusing to talk about the "data generating process".

jpaillard: I see your point. I changed the formulation.

bthirion (approved):
LGTM thx ! 
Maybe change the title to "Variable Selection under Model Mis-specification" (?)

lionelkusch (approved):
thank you.

PR #229, opened 2025-04-09
Add classification example on the iris dataset

jpaillard:
linked to #219


examples/plot_classification_example.py
```diff
@@ -0,0 +1,247 @@
...
+            "vim": vim_name,
+            "model": model_name,
+            "score": balanced_accuracy_score(y_true=y[test_index], y_pred=y_pred),
+        }
+    )
+
+
+models = [
+    LogisticRegressionCV(Cs=np.logspace(-3, 3, 10)),
+    GridSearchCV(LinearSVC(), {"C": np.logspace(-3, 3, 10)}),
```
bthirion, line 106: These models are very similar. I don't see the benefit of comparing them.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,247 @@
...
+        X, y, model, train_index, test_index, vim_name=vim_name, groups=groups
+    )
+    for train_index, test_index in cv.split(X)
+    for model in models
+    for vim_name in ["CPI", "LOCO", "PermutationImportance"]
+)
+df = pd.concat(out_list)
+
+
+def compute_pval(df, threshold=0.05):
```
bthirion, line 121: Would it be possible to avoid such functions and directly write loops in the examples ?.  This makes the examples harder to understand

bthirion (commented):
Thx for starting that.

examples/plot_classification_example.py
jpaillard: I wrote a function because the code is used twice (once for single variables and once for groups). Would you rather have the loop duplicated?

examples/plot_classification_example.py
jpaillard: I understand. My idea was to show how to change the prediction method (predict_proba & decision_function) and the losses (hinge and log_loss). I suggest changing to an SVC with rbf kernel. WDYT?

bthirion: Sounds reasonable

examples/plot_classification_example.py
bthirion: I see. I find this a bit frustrating for users because the interaction with the classes is burried in functions.
I'll try to make a suggestion upon my next review.

jpaillard: I am not sure to see what you are referring to. Is it the `compute_pval` function, which only computes a test and does data frame manipulations? Or is it the `run_one_fold`? 
IMO the latter is useful to show how to parallelize the computation of importance over K-folds.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,244 @@
...
+    if isinstance(model_c, LogisticRegressionCV):
+        method = "predict_proba"
+        loss = log_loss
+        model_name = "LogReg"
+    else:
+        method = "decision_function"
+        loss = hinge_loss
+        model_name = "SVC"
+
+    if vim_name == "CPI":
```
lionelkusch, line 63: It's better, for example, to put the creation of the object outside of the function and add a short description of the methods. Most people don't know CPI, LOCO and PFI.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,244 @@
...
+            n_permutations=50,
+            random_state=0,
+            method=method,
+            loss=loss,
+        )
+
+    vim.fit(X[train_index], y[train_index], groups=groups)
+    importance = vim.importance(X[test_index], y[test_index])["importance"]
+
+    return pd.DataFrame(
```
lionelkusch, line 90: Why do you use a data frame?

examples/plot_classification_example.py
```diff
@@ -0,0 +1,244 @@
...
+            "feature": groups.keys(),
+            "importance": importance,
+            "vim": vim_name,
+            "model": model_name,
+            "score": balanced_accuracy_score(y_true=y[test_index], y_pred=y_pred),
+        }
+    )
+
+
+models = [
```
lionelkusch, line 101: You should add a short comment about the model.

examples/plot_variable_importance_classification.py
```diff
@@ -0,0 +1,156 @@
...
+for train_index, test_index in cv.split(X):
+    X_train, X_test = X[train_index], X[test_index]
+    y_train, y_test = y[train_index], y[test_index]
+
+    model_linear_ = clone(model_linear)
+    model_linear_.fit(X_train, y_train)
+    model_svc_ = clone(model_svc)
+    model_svc_.fit(X_train, y_train)
+
+    cpi_linear = CPI(
```
lionelkusch, line 68: you should create the object before the loop with a short description of the method.

examples/plot_classification_example.py
lionelkusch: In my opinion, you should remove this function and create 2 examples: one for grouping variables and one without it. You can make references between them to include the comparison.

lionelkusch (commented):
Thank you for these examples but we need to clarify the message and consider that people who will read it don't know about what are the different methods of variable importance.

examples/plot_classification_example.py
jpaillard: I find it convenient for plotting

examples/plot_variable_importance_classification.py
jpaillard: The class's `__init__` takes a fitted model as an argument. It cannot be created before fitting the model, which is done in the loop.

jpaillard:
I see your points. The example is not meant to describe CPI, LOCO, and PFI; I think we should have other examples for that but rather a walkthrough of the API and 'how to': 1. measure importance for classification and 2. use groups instead of individual variables.

examples/plot_variable_importance_classification.py
lionelkusch: Good point.

lionelkusch: Good point.

lionelkusch: In my opinion, it's better to create multiple loops when you create an example and do a small number of calculi than a big loop with everything.

You can generate all the estimators before to instantiate CPI.

lionelkusch:
> I see your points. The example is not meant to describe CPI, LOCO, and PFI; I think we should have other examples for that but rather a walkthrough of the API and 'how to': 1. measure importance for classification and 2. use groups instead of individual variables.

It is always good to make a short recall of it (1 line with a redirection to the API).

examples/plot_classification_example.py
```diff
@@ -0,0 +1,244 @@
...
+        X, y, model, train_index, test_index, vim_name=vim_name, groups=groups
+    )
+    for train_index, test_index in cv.split(X)
+    for model in models
+    for vim_name in ["CPI", "LOCO", "PermutationImportance"]
+)
+
+df_grouped = pd.concat(out_list)
+df_pval = compute_pval(df_grouped, threshold=threshold)
+plot_results(df_grouped, df_pval)
```
lionelkusch, line 244: Why LOCO doesn't select the 2 groups at then end?

examples/plot_classification_example.py
jpaillard: as discussed below, it has to be in the loop since it requires the fitted model

examples/plot_classification_example.py
jpaillard: My explanation would be that LOCO has less statistical power because of the larger variance it introduces when refitting sub-models. When the sample size is small (Iris has 150 samples ), this leads larger p-values.

examples/plot_classification_example.py
lionelkusch: see my response below.

examples/plot_variable_importance_classification.py
```diff
@@ -0,0 +1,156 @@
...
+from sklearn.datasets import load_wine
+from sklearn.linear_model import LogisticRegressionCV, RidgeCV
+from sklearn.metrics import balanced_accuracy_score, hinge_loss, log_loss
+from sklearn.model_selection import KFold
+from sklearn.svm import SVC
+
+from hidimstat import CPI, LOCO, PermutationImportance, desparsified_lasso
+
+# %%
+#############################################################################
```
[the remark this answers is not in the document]
jpaillard: added comments in the introduction part

examples/plot_variable_importance_classification.py
jpaillard: Also, note that the code is encapsulated in a function. The idea is to show that the code is easy to parallelize.

jpaillard:
LOCO and CPI should asymptotically provide similar results but differ in this small sample size.
PFI can be used to measure importance in classification but is not robust to spurious features.

lionelkusch:
In this case, you should choose one between model between LOCO and CPI. 
And we perhaps made an example to show the difference between LOCO and CPI.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,255 @@
...
+
+dataset.feature_names = dataset.feature_names + ["spurious_feat"]
+
+
+############################################################################
+# Measure variable importance
+# --------------------------------------------------------------------------
+# Since all three methods compute variable importance as a loss difference, they
+# require a K-fold cross-fitting. Computing the importance for each fold is
+# embarassingly parallel, so we use joblib to parallelize the computation.
```
bthirion, line 59: # For this reason, we encapsulate the main computations in a function

bthirion (commented):
The example looks great. I still ahve 2 requests:
* as required by @lionelkusch , if the message is on groups vs single variable, I think it is better not to insist on the difference between CPI and LOCO
* as for the other example, split the code into smaller chunks and explain what you're doing. The examples are mini-tutorials.

examples/plot_classification_example.py
jpaillard: integrated in the text

jpaillard:
- As you suggested, I removed LOCO to streamline the example. 
 - I added more text blocks to split the code into smaller chunks and add more explanations

examples/plot_variable_importance_classification.py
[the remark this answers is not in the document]
jpaillard: I just realized that this file was pushed by error. I removed it.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,258 @@
+"""
+Measuring Individual and Group Variable Importance for Classification
+======================================================================
+In this example, we show on the Iris dataset how to measure variable importance for
+classification tasks. We use two different variable importance methods: CPI and
+PermutationImportance with two different classifiers: LogisticRegressionCV and
```
lionelkusch, line 6:
```suggestion
classification tasks. We use two different variable importance methods: Conditional Permutation importance (CPI) and
Permutation Feature Importance (PFI) with two different classifiers: LogisticRegressionCV and
```

examples/plot_classification_example.py
```diff
@@ -0,0 +1,258 @@
...
+LinearSVC. We start by measuring the importance of individual variables and then show
+how to measure the importance of groups of variables.
+To briefly summarize the two methods:
+
+ - PFI (Permutation Feature Importance) shuffles the values of a feature and measures
+ the increase in the loss when predicting (using om the same full model) on the
+ shuffled data.
+ - CPI (Conditional Permutation Importance) is a conditional version of PFI that
```
lionelkusch, line 14: The rendering of the bullet point is not correct.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,258 @@
...
+
+from hidimstat import CPI, PermutationImportance
+
+########################################################################
+# Load the iris dataset and add a spurious feature
+# ----------------------------------------------------------------------
+# We load the iris dataset and add a spurious feature that is a linear combination of
+# the petal length, width amd some noise but not related to the target. The spurious feature
+# allows to illustrate that `PermutationImportance` is not robust to spurious features,
+# contrarily to `LOCO`` and `CPI``.
```
lionelkusch, line 40:
```suggestion
# allows to illustrate that `PFI` is not robust to spurious features,
# contrarily to `CPI``.
```

examples/plot_classification_example.py
lionelkusch, line 6: In the explanation use the common name of the model and not the name of the function.
Can you replace `LogisticRegressionCV` by `Logistic Regression` and `LinearSVC` by `Linear Support Vector Classification`?

lionelkusch: Same remark for `Permutation Importance` use `PFI`.

examples/plot_classification_example.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+plot_results(df, df_pval)
+
+
+####################################################################################
+# The boxplot shows the importance of each feature, with colors indicating the
+# classifier used. A star marks the features that have a p-value (computed with a
+# t-test) below 0.05. As expected, the spurious feature is not selected by CPI,
+# but is selected by Permutation Importance.
```
lionelkusch, line 233: Can you add a comment between LogReg and SVC model?

examples/plot_classification_example.py
jpaillard: done

examples/plot_classification_example.py
jpaillard: done

lionelkusch (commented):
Can you check  for removing the following warning?
"
/home/circleci/project/doc_conf/auto_examples/plot_classification_example.rst:29: ERROR: Unexpected indentation. [docutils]
/home/circleci/project/doc_conf/auto_examples/plot_classification_example.rst:30: WARNING: Bullet list ends without a blank line; unexpected unindent. [docutils]
/home/circleci/project/doc_conf/auto_examples/plot_classification_example.rst:34: WARNING: Bullet list ends without a blank line; unexpected unindent. [docutils]
"

examples/plot_classification_example.py
bthirion: I get the point. (and I think that this pattern is also used in sklearn, so...)

bthirion (approved):
LGTM, thx.

lionelkusch:
There is this warning on this new example:
```
Increase the number of iterations (max_iter) or scale the data as shown in:
    https://scikit-learn.org/stable/modules/preprocessing.html
Please also refer to the documentation for alternative solver options:
    https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression
  n_iter_i = _check_optimize_result(
/home/circleci/project/miniconda/envs/testenv/lib/python3.12/site-packages/sklearn/linear_model/_logistic.py:465: ConvergenceWarning: lbfgs failed to converge (status=1):
STOP: TOTAL NO. OF ITERATIONS REACHED LIMIT.
```

Do you think that you can remove it or can you filter the warning to ignore it?

After this it will be good.

After this it will be good.

lionelkusch:
@jpaillard Tell me if my modification is OK for you.

jpaillard:
It's good on my side. I let you merge if it's good for you @lionelkusch

PR #233, opened 2025-04-14
Remove example based on complex simulated data

jpaillard:
This PR is the last step of #219. It removes the existing example: "Measuring variable importance in classification" which uses unusual simulated data. 
All the aspects of the library showcased in this example have been transferred to other examples.


bthirion:
Good with me to remove it. @jpaillard since the effort was yours, let us know whether we overlooked some interesting aspect that would justify keeping it.

PR #216, opened 2025-04-04
Knockoff example

alexblnn:
Adding knockoff example

PR #236, opened 2025-04-15
Knockoffs example - follow-up

jpaillard:
Following up on #216, started by @alexblnn. 

I suggest minor modifications and a visual representation & comparison of the selection sets for KO and Lasso.


examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,235 @@
...
+seed = 0
+rng = np.random.RandomState(seed)
+
+
+########################################################################################
+# Load the breast cancer dataset
+# ------------------------------------------------------
+# There are 569 samples and 30 features that correspond to tumor attributes.
+# The downstream task is to classify tumors as benign or malignant. We leave out 10% of
+# the data to evaluate the performance of the Lasso model on the prediction task.
```
bthirion, line 25: Note that it's not a lasso strictly speaking. Maybe call it "Logistic Lasso" or "sparse logistic".

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,235 @@
...
+        "selected": selected_mask,
+    }
+)
+# Count how many selected features are actually noise
+num_false_discoveries = np.sum(selected_logl1 >= p)
+print(f"The Lasso makes at least {num_false_discoveries} False Discoveries!!")
+
+
+#################################################################################
+# The Lasso selects many spurious variables that are not related to the outcome.
```
bthirion, line 124:
```suggestion
# The Lasso selects many spurious variables that are not directly related to the outcome.
```

bthirion (commented):
The example looks great overall.

examples/plot_knockoffs_wisconsin.py
jpaillard: I tried to clarify this.

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,236 @@
...
+lasso_noisy = LogisticRegressionCV(
+    Cs=np.logspace(-3, 3, 10), penalty="l1", solver="liblinear", random_state=rng
+)
+lasso_noisy.fit(noisy_train, y_train)
+y_pred_noisy = lasso_noisy.predict(noisy_test)
+print(
+    f"Accuracy of Lasso on test set with noise: {lasso_noisy.score(noisy_test, y_test):.3f}"
+)
+
+selected_logl1 = np.where(np.abs(lasso_noisy.coef_[0]) > 1e-6)[0]
```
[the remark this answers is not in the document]
jpaillard: done

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+#############################################################################
+# Selecting variables with the Lasso
+# ------------------------------------------
+# We want to select variables that are relevant to the outcome, i.e. tumor
+# charateristics that are associated with tumor malignance. We start off by applying a
+# classical method using Lasso logistic regression and retaining variables with non-null
+# coefficients:
+from sklearn.linear_model import LogisticRegressionCV
```
lionelkusch, line 53: Why do you import LogisticRegressionCV and not Lasso?

You should add an short explanation in the text.

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,237 @@
...
+        solver="liblinear",
+        penalty="l1",
+        Cs=np.logspace(-3, 3, 10),
+        random_state=rng,
+    ),
+    n_bootstraps=1,
+    random_state=0,
+    preconfigure_estimator=None,
+    fdr=fdr,
+    n_jobs=3,
```
lionelkusch, line 155: The definition of n_jobs is unnecessary in this context.

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,237 @@
...
+
+noisy_train = np.concatenate(noises_train, axis=1)
+noisy_test = np.concatenate(noises_test, axis=1)
+
+
+#################################################################################
+# There are 180 features, 30 of them are real and 150 of them are fake and independent
+# of the outcome. We now apply the Lasso (with cross-validation to select the best
+# regularization parameter) to the noisy dataset and observe the results:
+lasso_noisy = LogisticRegressionCV(
```
lionelkusch, line 99: The number of jobs should be defined for this class.

examples/plot_knockoffs_wisconsin.py
```diff
@@ -0,0 +1,237 @@
...
+# Controlled variable selection with Knockoffs
+# ----------------------------------------------------
+# We use the Model-X Knockoff procedure to control the FDR (False Discovery Rate). The
+# selection of variables is based on the Lasso Coefficient Difference (LCD) statistic
+# :footcite:t:`candes2018panning`.
+from hidimstat import model_x_knockoff
+
+fdr = 0.2
+
+selected, test_scores, threshold, X_tildes = model_x_knockoff(
```
lionelkusch, line 142: There are some Convergence Warnings. Can you fix them or ignore them?

examples/plot_knockoffs_wisconsin.py
jpaillard: It is because Lasso is a regression model while the Wisconsin dataset is a binary classification problem. Following the previous comment from @bthirion I modified the text to include: 
 > [...] the Logistic Lasso (Logistic Regression with L1 regularization) on the prediction task.

examples/plot_knockoffs_wisconsin.py
jpaillard: I removed `n_jobs`, it's not critical in this light example.

examples/plot_knockoffs_wisconsin.py
jpaillard: done

examples/plot_knockoffs_wisconsin.py
lionelkusch: Can you also add it in the section name?

examples/plot_knockoffs_wisconsin.py
lionelkusch: You should put it at 1.
I want to have control over the different n_jobs in the examples.

examples/plot_knockoffs_wisconsin.py
lionelkusch: in this case, the n_job is define by the number of bootstrap.

examples/plot_knockoffs_wisconsin.py
jpaillard: done

bthirion (approved):
LGTM, thx !

lionelkusch (commented):
Can you look to remove this warning?

/home/circleci/project/miniconda/envs/testenv/lib/python3.12/site-packages/sklearn/utils/validation.py:1408: DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples, ), for example using ravel().
  y = column_or_1d(y, warn=True)

lionelkusch (approved):
> The warning messages seem to come from `knockoff_aggregation` and `diabetes_variable_importance_example`. I suggest fixing them in a separate PR.
Ok, in this case it is fine.
For the `knockoff_aggregation`, this will be fixed in #215.
We will need an additional PR for `diabetes_variable_importance_example`

alexblnn:
Thanks a lot @jpaillard, looks great! A small detail on the figure at the end: perhaps don't use "rejected" and "selected". Selected variables are often returned as a "rejection set" by multiple testing procedures (BH etc) so we should avoid this confusion. I believe you mean "selected" and "not selected" which is right IMO.

jpaillard:
I see, I updated the figure's legend accordingly. Thx for the review.

PR #342, opened 2025-08-21
Replace `n_alphas` by `alphas` for sklearn compatibility

jpaillard:
This should remove the warning:
> warnings.warn(
/home/circleci/project/.venv/lib/python3.13/site-packages/sklearn/linear_model/_coordinate_descent.py:1622: FutureWarning: 'n_alphas' was deprecated in 1.7 and will be removed in 1.9. 'alphas' now accepts an integer value which removes the need to pass 'n_alphas'. The default value of 'alphas' will change from None to 100 in 1.9. Pass an explicit value to 'alphas' and leave 'n_alphas' to its default value to silence this warning.

### DIFF
--- a/src/hidimstat/distilled_conditional_randomization_test.py
+++ b/src/hidimstat/distilled_conditional_randomization_test.py
@@ -1,15 +1,16 @@
-import numpy as np
 import warnings
+
+import numpy as np
 from joblib import Parallel, delayed
 from scipy import stats
 from sklearn.base import clone
 from sklearn.linear_model import Lasso, LassoCV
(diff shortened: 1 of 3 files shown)


jpaillard:
According to scikit-learn's documentation: 
> Deprecated since version 1.7: n_alphas was deprecated in 1.7 and will be removed in 1.9. Use alphas instead.

Making this change will imply updating the min version of sklearn to 1.7 (see failed tests minimal). Do we want to make this update?

jpaillard:
If we don't want to comply to the latest sklearn version, should I simply close this PR?

PR #370, opened 2025-09-02
`D0CRT` Accept Scikit-learn models directly instead of parameter dictionaries

jpaillard:


src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -69,25 +61,27 @@ class D0CRT(BaseVariableImportance):
...
+        Whether to reuse the screening model for y-distillation.
 
     Attributes
     ----------
     coefficient_ : ndarray of shape (n_features,)
         Estimated feature coefficients after screening/refitting during fit method.
-    clf_x_ : list of estimators of length n_features
+    selection_set : ndarray of shape (n_features,)
+        Boolean mask indicating selected features after screening.
+    model_x_ : list of estimators
```
jpaillard, line 73: renamed `clf` --> `model` since it is typically a regression model.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -69,25 +61,27 @@ class D0CRT(BaseVariableImportance):
         Whether to use scaled statistics when computing importance.
     random_state : int, default=2022
         Random seed for reproducibility.
+    reuse_screening_model: bool, default=True
```
jpaillard, line 64: From what I understand this was hidden, ie, the screening weights were used when screening=True

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -31,28 +32,19 @@ class D0CRT(BaseVariableImportance):
...
-        - selection : {'cyclic'} - Feature selection method (default: 'cyclic').
-    params_lasso_distillation_x : dict or None, default=None
-        Parameters for X distillation Lasso. If None, uses params_lasso_screening.
+    lasso_screening : sklearn estimator
+        Estimator for variable screening (typically LassoCV or Lasso).
+    model_distillation_x : sklearn estimator
+        Estimator for X distillation (typically LassoCV or Lasso).
     refit : bool, default=False
         Whether to refit the model on selected features after screening.
-    screening : bool, default=True
```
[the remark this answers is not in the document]
jpaillard: My argument is that it simplifies the code and avoids passing conflicting arguments. The computation remains the same.

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: If a non-linear model is used for y-distillation, and screening is performed.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -31,28 +31,19 @@ class D0CRT(BaseVariableImportance):
         The base estimator used for y-distillation and prediction
         (e.g., Lasso, RandomForest, ...).
     method : str, default="predict"
-        Method of the estimator to use for predictions ("predict", "predict_proba", etc.).
+        Method of the estimator to use for predictions ("predict", "predict_proba",
```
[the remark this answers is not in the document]
lionelkusch: Yes, it can be `decision_function` (see https://scikit-learn.org/dev/developers/develop.html#different-objects)

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -439,90 +401,71 @@ def fit_importance(self, X, y, cv=None):
...
+         Index of the variable to be tested.
+     X : array-like of shape (n_samples, n_features)
+         The input data matrix.
+     y : array-like of shape (n_samples,)
+         The target values.
+     estimator : sklearn estimator
+         The estimator used for distillation and prediction.
+     sigma_X : bool, default=False
+         If True, use Lasso for X distillation; if False, skip X distillation.
+     fit_y : bool, default=False
```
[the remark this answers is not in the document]
lionelkusch: Can you update the doctring?

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -683,27 +598,25 @@ def d0crt(
...
-    screening_threshold=1e-1,
+    screening_threshold=10,
     centered=True,
     n_jobs=1,
     joblib_verbose=0,
     fit_y=False,
-    random_state=2022,
     scaled_statistics=False,
+    random_state=2022,
+    reuse_screening_model=True,
```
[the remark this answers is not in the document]
lionelkusch: The docstring is generated base on the class function.

jpaillard:
Would it actually make sense to have a `Screening` class instead of an independent function? 
The class could not only be compatible with Lasso but also the univariate methods from #317

src/hidimstat/distilled_conditional_randomization_test.py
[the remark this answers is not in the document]
jpaillard: done

src/hidimstat/distilled_conditional_randomization_test.py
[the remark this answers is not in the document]
lionelkusch: In reality, there are 4 possibilities in this context: "predict", "predict_proba", "decision_function", "transform".
See the proposition for CFI:
https://github.com/mind-inria/hidimstat/blob/bafa4ed05412dc077fea700f4dc82bac5505cc05/src/hidimstat/conditional_feature_importance.py#L36-L39

examples/plot_model_agnostic_importance.py
```diff
@@ -162,8 +162,9 @@
 # As expected, when using linear models (d0CRT and LOCO-linear) that are misspecified,
 # the varibles are not selected. This highlights the benefit of using model-agnostic
 # methods such as LOCO, which allows for the use of models that are expressive enough
-# to explain the data.
-
+# to explain the data. While d0CRT can use any estimator, its distillation step
+# restricts it from capturing variable interactions. A more advanced distillation
+# approach, such as d1CRT, may help overcome this limitation.
```
bthirion, line 167: If we provide no implementation of d1CRT, we should not make any claim on it.

src/hidimstat/distilled_conditional_randomization_test.py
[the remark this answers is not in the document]
bthirion: "transform" sounds really weird to me in this context. Is there any instance where it is used (tests or examples) ? If not we should not include it.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -64,30 +55,38 @@ class D0CRT(BaseVariableImportance):
     joblib_verbose : int, default=0
         Verbosity level for parallel jobs.
     fit_y : bool, default=False
-        Whether to fit y using selected features instead of using estimated_coef.
+        Controls y-distillation behavior:
+        - If False and the estimator is linear, the sub-model predicting y from X^{-j}
+        is created by simply removing the idx-th coefficient from the full model
```
bthirion, line 60: Is this proceudre somehow validated. It sounds wrong as soon as there is correlation btw variables.

src/hidimstat/distilled_conditional_randomization_test.py
bthirion: But the screening seems to be based on (expensive to compute) Lasso coefficients... It would be a bit silly to do screening with a Lasso and then used another learner. But I agree that a separate issue/PR can be opened for that.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -109,44 +105,35 @@ def __init__(
...
     ):
         self.estimator = estimator
         _check_vim_predict_method(method)
         self.estimated_coef = estimated_coef
         self.method = method
         self.sigma_X = sigma_X
-        self.params_lasso_screening = params_lasso_screening
-        self.params_lasso_distillation_x = params_lasso_distillation_x
+        self.lasso_screening = lasso_screening
+        self.model_distillation_x = model_distillation_x
```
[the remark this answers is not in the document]
bthirion: Yep, see my comment above --- please an open an issue for this.

test/test_distilled_conditional_randomization_test.py
```diff
@@ -347,10 +324,84 @@ def test_warning_not_used_parameters():
         _ = d0crt.fit_importance(X, y, cv=cv)
 
 
+def test_dcrt_invalid_lasso_screening(generate_regation_dataset):
+    """
+    Test that passing a non-Lasso model to lasso_screening raises a ValueError.
+    """
+    X, y = generate_regation_dataset
```
bthirion, line 331: regation?

bthirion (commented):
Almost ready to merge, thx !

src/hidimstat/distilled_conditional_randomization_test.py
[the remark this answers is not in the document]
jpaillard: ok, done

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: I don't think so. Looking at the original R implementation by moleibobliu, it seems that it is not allowed: https://github.com/moleibobliu/Distillation-CRT/blob/dddc8cc4027c236233cb9f0afb65290fb622be15/functions.R#L155-L158
So I guess this option was introduced by @achamma723 when implementing d0CRT for the NeurIPS paper? 

Maybe we could put a warning while trying to figure out why it was introduced?

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: I opened the issue #378

test/test_distilled_conditional_randomization_test.py
jpaillard: typo: --> regression

src/hidimstat/distilled_conditional_randomization_test.py
lionelkusch: see PR #380

src/hidimstat/distilled_conditional_randomization_test.py
bthirion: What would be useful would be to check that this behavior gives sound results (e.g. that refitting would yield the same results empirically).

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -98,6 +97,9 @@ class D0CRT(BaseVariableImportance):
     3. Test statistic computation using residual correlations.
     4. P-value calculation assuming Gaussian null distribution.
 
+    The implementation currently allows flexible model for the y-distillation step.
```
bthirion, line 100:
```suggestion
    The implementation currently allows flexible models for the y-distillation step.
```

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -683,27 +612,25 @@ def d0crt(
...
-    screening=True,
-    screening_threshold=1e-1,
+    screening_threshold=10,
     centered=True,
     n_jobs=1,
     joblib_verbose=0,
     fit_y=False,
-    random_state=2022,
     scaled_statistics=False,
+    random_state=2022,
```
bthirion, line 632: We should not set random_state inside the function.

test/test_distilled_conditional_randomization_test.py
```diff
@@ -2,38 +2,37 @@
...
     return X, y
 
 
-def test_dcrt_lasso_screening(generate_regation_dataset):
+def test_dcrt_lasso_screening(generate_reression_dataset):
     """
     Test for screening parameter and pvalue function
     """
-    X, y = generate_regation_dataset
+    X, y = generate_reression_dataset
```
bthirion, line 27:
```suggestion
    X, y = generate_regression_dataset
```

test/test_distilled_conditional_randomization_test.py
```diff
@@ -61,19 +60,19 @@ def test_dcrt_lasso_screening(generate_regation_dataset):
     assert len(d0crt_no_screening.importances_) == 10
 
 
-def test_dcrt_lasso_with_estimed_coefficient(generate_regation_dataset):
+def test_dcrt_lasso_with_estimed_coefficient(generate_reression_dataset):
     """
     Test the estimated coefficient parameter
     """
-    X, y = generate_regation_dataset
+    X, y = generate_reression_dataset
```
bthirion, line 67:
```suggestion
    X, y = generate_regression_dataset
```

test/test_distilled_conditional_randomization_test.py
```diff
@@ -2,38 +2,37 @@
...
-
-from hidimstat import d0crt, D0CRT
+from hidimstat import D0CRT, d0crt
 from hidimstat._utils.regression import _alpha_max
 
 
 @pytest.fixture
-def generate_regation_dataset(n=100, p=10, noise=0.2, seed=2024):
-    X, y = make_regression(n_samples=n, n_features=10, noise=0.2, random_state=2024)
+def generate_reression_dataset(n=100, p=10, noise=0.2, seed=2024):
```
bthirion, line 18:
```suggestion
def generate_regression_dataset(n=100, p=10, noise=0.2, seed=2024):
```

bthirion (commented):
Almost there, thx !

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: A quick check using make_regression shows very similar results: 
image

The full model is a lasso trained with 10 features, then the 10th coefficient is deleted. Sub-model is a model fitted on 9 features. The code used to generate it is the following: 
```python
# %%
import numpy as np
import pandas as pd
import seaborn as sns
from sklearn.datasets import make_regression
from sklearn.linear_model import LassoCV
from sklearn.model_selection import KFold

# %%
res_list = []
seed = 0
rng = np.random.RandomState(seed)
X, y = make_regression(n_samples=1000, n_features=10, noise=0.5, random_state=seed)
var_idx = rng.random_integers(0, X.shape[1] - 1)
cv = KFold(n_splits=5, shuffle=True, random_state=seed)
for i, (train, _) in enumerate(cv.split(X)):
    X_train, y_train = X[train], y[train]
    model = LassoCV(random_state=seed)
    model.fit(X_train, y_train)
    coef_full = np.delete(model.coef_, var_idx)
    res_list.append(
        pd.DataFrame(
            {
                "model": "full model",
                "coef_val": coef_full,
                "fold": i,
                "coef_id": np.arange(len(coef_full)),
            }
        )
    )

    sub_model = LassoCV(random_state=seed)
    sub_model.fit(np.delete(X_train, var_idx, axis=1), y_train)
    coef_sub = sub_model.coef_
    res_list.append(
        pd.DataFrame(
            {
                "model": "sub model",
                "coef_val": coef_sub,
                "fold": i,
                "coef_id": np.arange(len(coef_sub)),
            }
        )
    )
df_plot = pd.concat(res_list, ignore_index=True)
sns.boxplot(data=df_plot, x="coef_id", y="coef_val", hue="model")
```

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: I set the default to None

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: My bad. Indeed, as expected generating the data with correlation makes a big difference. 
```python
corr_matrix = toeplitz(0.5 ** np.arange(10))
X = rng.multivariate_normal(mean=np.zeros(10), cov=corr_matrix, size=1000)
beta = rng.randn(10)
y = X @ beta
y += rng.normal(0, np.std(y) / snr, size=y.shape)
```

jpaillard: IMO, we should remove this possibility, WDYT?. 
Should I tackle that in this PR or open an issue with a proper test showing the problematic behavior and then fix it?

jpaillard: I opened #383

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -109,44 +111,35 @@ def __init__(
...
         self.refit = refit
-        self.screening = screening
         self.screening_threshold = screening_threshold
         self.centered = centered
         self.n_jobs = n_jobs
         self.joblib_verbose = joblib_verbose
         self.fit_y = fit_y
         self.scaled_statistics = scaled_statistics
         self.random_state = random_state
+        self.is_lasso = isinstance(estimator, (Lasso, LassoCV))
```
lionelkusch, line 141: Why do you need this attribute?
It seems only used once.

lionelkusch: If you want to keep it, you should add an underline to this end and put it after the attributes defined by the parameters.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -109,44 +111,35 @@ def __init__(
...
+        self.reuse_screening_model = reuse_screening_model
```
lionelkusch, line 142:
```suggestion
        self.reuse_screening_model = reuse_screening_model
        
        self.lasso_model_ = None
        self.coefficient_ = None
```

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -191,83 +184,59 @@ def fit(self, X, y):
...
+            self.coefficient_[~self.selection_set] = 0
         else:
-            self.coefficient_ = None
-            self.non_selection_ = np.zeros_like(np.arange(n_features), dtype=bool)
-            selection_set = np.logical_not(self.non_selection_)
+            self.estimator.fit(X_[:, self.selection_set], y_)
+            if hasattr(self.estimator, "coef_"):
+                self.coefficient_ = self.estimator.coef_
+                # optimisation to reduce the number of elements different to zeros
+                # TODO: the multi-dimensional indexing should be removed when the
```
lionelkusch, line 219: You should open an issue on this topic.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -191,83 +184,59 @@ def fit(self, X, y):
...
-                if hasattr(self.clf_screening_, "alpha_"):
-                    alpha_screening = self.clf_screening_.alpha_
-                else:
-                    alpha_screening = self.clf_screening_.alpha
-                # update the alpha value from the estimator on all values
-                self.params_lasso_screening["alpha"] = alpha_screening
-                if self.params_lasso_distillation_x is not None:
-                    self.params_lasso_distillation_x["alpha"] = alpha_screening
-                self.coefficient_ = np.ravel(self.clf_screening_.coef_)
+                self.selection_set = np.ones(X.shape[1], dtype=bool)
```
lionelkusch, line 194:
```suggestion
                self.selection_set_ = np.ones(X.shape[1], dtype=bool)
```
`selection_set_` is an generated attribute; you should add an underline at this end. 
You should rename it in all the files.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -346,18 +315,19 @@ def importance(
         else:
             X_ = X
         n_samples, n_features = X_.shape
-        selection_features = np.arange(n_features)[np.logical_not(self.non_selection_)]
+        selection_features = np.arange(n_features)[self.selection_set]
```
lionelkusch, line 318:
```suggestion
```
It's not necessary to create a variable for it.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -346,18 +315,19 @@ def importance(
...
 
         ## Distillation & calculate
         list_job = []
-        for index, (idx, clf_y, clf_x) in enumerate(
-            zip(selection_features, self.clf_y_, self.clf_x_)
+        for index, (idx, model_y, model_x) in enumerate(
+            zip(selection_features, self.model_y_, self.model_x_)
```
lionelkusch, line 323:
```suggestion
            zip(np.where(self.selection_set)[0], self.model_y_, self.model_x_)
```

lionelkusch: Due to my proposition of removing `selection_features`.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -346,18 +315,19 @@ def importance(
...
         ):
-            if self.coefficient_ is not None:
+            if self.is_lasso:
```
lionelkusch, line 325: Why is it only for lasso?

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -499,38 +460,34 @@ def _joblib_fit(
...
-            n_jobs=1,
-            **params_lasso_distillation_x,
-        )
+        model_x = clone(model_distillation_x)
+        model_x.fit(X_minus_idx, X[:, idx])
     else:
-        clf_x = None
+        model_x = None
 
     if (isinstance(estimator, Lasso) or isinstance(estimator, LassoCV)) and not fit_y:
```
lionelkusch, line 468: I think we need to check if it's a linear model instead than if it's lasso.
What do you think?

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -610,69 +567,41 @@ def _joblib_distill(
...
+def run_lasso_screening(
     X,
     y,
-    n_jobs,
-    alpha,
-    alphas,
-    n_alphas,
-    alpha_max_fraction,
-    **params,
+    lasso_model=LassoCV(fit_intercept=False),
```
lionelkusch, line 573:
```suggestion
    lasso_model=LassoCV(fit_intercept=False, random_state=0),
```

For default parameters, I prefer that the random_state is fixed.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -610,69 +567,41 @@ def _joblib_distill(
...
-        alphas are set automatically.
-    n_alphas : int
-        Number of alphas along the regularization path. Ignored if alphas is provided.
-    alpha_max_fraction : float
-        Fraction of alpha_max to use when alpha, alphas, and n_alphas are not provided.
-    random_state : int, RandomState instance or None
-        Random seed for reproducibility.
-    **params : dict
-        Additional parameters for Lasso or LassoCV.
+    lasso_model : sklearn estimator or None, default=LassoCV(fit_intercept=False)
```
lionelkusch, line 586:
```suggestion
    lasso_model : sklearn estimator (uniquely Lasso or LassoCV) or None, default=LassoCV(fit_intercept=False)
```

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -109,44 +111,35 @@ def __init__(
...
-            "alpha_max_fraction": 0.5,
-            "cv": 5,
-            "tol": 1e-6,
-            "max_iter": 1000,
-            "fit_intercept": False,
-            "selection": "cyclic",
-        },
-        params_lasso_distillation_x=None,
+        lasso_screening=LassoCV(n_alphas=10, cv=5, tol=1e-6, fit_intercept=False),
+        model_distillation_x=LassoCV(n_jobs=1, n_alphas=10),
```
lionelkusch, line 115:
```suggestion
        lasso_screening=LassoCV(n_alphas=10, cv=5, tol=1e-6, fit_intercept=False, random_state=0),
        model_distillation_x=LassoCV(n_jobs=1, n_alphas=10, random_state=0),
```
I prefer to fix the random generator for the default parameters.

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -683,27 +612,25 @@ def d0crt(
...
+    lasso_screening=LassoCV(
+        n_alphas=10,
+        cv=5,
+        tol=1e-6,
+        fit_intercept=False,
+    ),
+    model_distillation_x=LassoCV(
+        n_jobs=1,
+        n_alphas=10,
+    ),
```
lionelkusch, line 624:
```suggestion
    lasso_screening=LassoCV(
        n_alphas=10,
        cv=5,
        tol=1e-6,
        fit_intercept=False,
         random_state=0,
    ),
    model_distillation_x=LassoCV(
        n_jobs=1,
        n_alphas=10,
        random_state=0,
    ),
```

test/test_distilled_conditional_randomization_test.py
```diff
@@ -190,18 +183,14 @@ def test_dcrt_lasso_refit():
...
     d0crt = D0CRT(
         estimator=LassoCV(n_jobs=1), estimated_coef=np.ones(10) * 10, random_state=2024
     )
     d0crt.fit(X, y)
-    for arr in [
-        d0crt.clf_x_,
-        d0crt.clf_y_,
-    ]:
-        assert np.all(arr == np.array([]))
+    assert np.all(d0crt.selection_set)
```
lionelkusch, line 193: Why do you change this test?

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: I will remove it, it will not be used in the future after taking care of #383

src/hidimstat/distilled_conditional_randomization_test.py
jpaillard: It's just convenient since it's used twice L323 and L371

src/hidimstat/distilled_conditional_randomization_test.py
lionelkusch: see above

src/hidimstat/distilled_conditional_randomization_test.py
lionelkusch: No, Lasso is also a source of randomness. 
Lasso in scikitlearn requires a random generator as a parameter.

lionelkusch: In your actual modification, I count 5 random generators to define for having reproducible results:
- 2 for the cross-validation generator of the LassoCV
- 2 for the model of lasso
- 1 for the method

For the simplicity of handling the random generator, I prefer that all the other random generators as a fixed seed except the random generator of the method.

jpaillard: Yes, but the generator of the method is actually never used. I suggest fixing this in #368 
Consequently, if you fix the seed of the models, there is no more randomness by default, which I think should be the default behavior. 

Let me know what you think.

jpaillard: OK

Issue #306, opened 2025-07-01
Create narrative documentation


bthirion:
I just realized that Hidimstat has no narrative documentation. It is clearly the main missing piece atm. This something we need to work on.
It has to be tightly coupled with examples as Nilearn and sklearn. Adding stuff on pdp, marginal vim etc. will be important in that respect.
It might be organized as follows:
* Definition of variable importance
** variable importance, variable selection
** conditional vs marginal variable importance
** roadblocks toward variable importance computation: dimensionality and correlation
** statistical inference / confidence intervals
* A review of vi methods, organized according to the target criterions
* Very high dimensional case (EncludL etc)

lionelkusch:
If I am correct, it is missing a user guide.

I made a selection of user guides for getting inspiration:
- nilearn: https://nilearn.github.io/stable/user_guide.html#user-guide
- scikit-learn: https://scikit-learn.org/stable/user_guide.html
- skrub: https://skrub-data.org/stable/documentation.html
- PIML: https://selfexplainml.github.io/PiML-Toolbox/_build/html/user_guide.html
- Nest simulator: https://nest-simulator.readthedocs.io/en/stable/get-started_index.html
- Neuron simulator: https://nrn.readthedocs.io/en/8.2.7/guide/index.html

Some more related to Feature Importance:
- Feature-engine: https://feature-engine.trainindata.com/en/latest/user_guide/selection/index.html
- Alibi: https://docs.seldon.io/projects/alibi/en/stable/overview/high_level.html
- SHAP: https://shap.readthedocs.io/en/latest/example_notebooks/overviews/An%20introduction%20to%20explainable%20AI%20with%20Shapley%20values.html
- VIP: https://koalaverse.github.io/vip/articles/vip.html

bthirion:
Outline V2:

-  Definition of variable importance
    -  variable importance, variable selection
    - conditional vs marginal variable importance
        - introduce Total Sobol Index
    - roadblocks toward variable importance computation: dimensionality and correlation
    - statistical inference / confidence intervals
- A review of vi methods, organized according to the target criterion
    -  Conditional inference
        - GLM-based inference: DL, MX-knockoffs, CRT, dCRT
        - Model agnostic methods that estimate TSI: LOCO, CPI
        - Permutation importance
     - marginal inference: f-regression, f-classif 
- Plotting utilities 
- Very high dimensional case (EncludL etc)

PR #405, opened 2025-09-15
[DOC] Add FI visualization & partial dependence plot to user guide

GaetandeCast:
Towards #306

Write a section for the visualization section of the user guide.

based on #318

### DIFF
--- a/docs/src/methods_list.rst
+++ b/docs/src/methods_list.rst
@@ -1,6 +1,0 @@
-.. _methods_list:
-
-
-===============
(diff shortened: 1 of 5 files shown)


lionelkusch:
> I guess we first need to merge #318 because the diff is a bit terrible atm.

yes.

GaetandeCast:
Hi @jpaillard, unfortunately I don't think I will have much time in the following weeks to come back to this. Feel free to take over.

PR #413, opened 2025-09-15
[DOC] move subfunction closer to where they are used

Remi-Gau:
- closes none
- move subfunction closer to where they are used in examples/plot_2D_simulation_example.py


bthirion (approved):
I'm happy  with the PR, but adding relatively long functions is a disaster. Maybe  we should consider having some of them in the code base.
Maybe the function is simply awkward and could be replaced by 2 lines of scipy (it looks like an image dilation does the job...).

Remi-Gau:
> Maybe we should consider having some of them in the code base.

agreed though as you said yesterday, the pattern of what "plotting tools" we need may show up as we have more examples

also I prefered to just move the functions first, before maybe inlining them

Issue #356, opened 2025-08-27
Nested joblib parallel loops in the examples generation


jpaillard:
The generation of the examples in the Sphinx doc is currently parallelized using joblib. However, some of our examples also use joblib for parallelization. This leads to nested parallel loops, which is bad for performance and might lead to reproducibility issues. 

I suggest removing the parallelization at the Sphinx doc generation level. Hidimstat examples are not too long, and there are not so many of them. In addition, it is important to illustrate in the example how the library's code can be parallelized and how/when to use `n_jobs` argument.

PR #416, opened 2025-09-15
[DOC] use consistent divider and header length

Remi-Gau:
- uses  `# %%` as example dividers (simpler and shorter than repeats of an undetermined number of `#`)
- clean up some 'header' length that were longer than the title

### DIFF
--- a/examples/plot_2D_simulation_example.py
+++ b/examples/plot_2D_simulation_example.py
@@ -48,7 +48,7 @@ References
 """
 import matplotlib.pyplot as plt
 
-#############################################################################
+# %%
 # Imports needed for this script
 # ------------------------------
 import numpy as np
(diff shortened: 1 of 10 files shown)


jpaillard (approved):
Looks better. 
The tests are surprisingly crashing, probably a bug. Is there a way to relaunch them?

bthirion (approved):
LGTM, thx.

lionelkusch:
> Looks better. The tests are surprisingly crashing, probably a bug. Is there a way to relaunch them?

@jpaillard the test minimal sometimes crashes because it doesn't correctly manage the parallelisation of the tests sometimes.
Once all the GitHub jobs are finished, it's possible to run only failed jobs in the GitHub workflow.

jpaillard:
Maybe linked to #356, we should still think about avoiding the parallel generation of examples since we have parallelization inside examples that we would like to keep.

lionelkusch:
this is the parallelization of the tests and not the examples.

PR #419, opened 2025-09-15
[DOC] add a glossary and notations page

jpaillard:


jpaillard:
I added LOCO, CFI, and PFI to the glossary, as well as an example showing how to refer to terms of the glossary guide in examples/doc. 
Would that make sense to merge this without waiting for it to be exhaustive, so that the other open PRs can build on top of it and include new terms?

docs/src/glossary_and_notations.rst
```diff
@@ -0,0 +1,60 @@
...
+===========================
+Glossary and Notations
+===========================
+
+Notations
+---------
+
+.. glossary::
+
+    minus index
```
lionelkusch, line 13: Can you add the definition of `minus index`?

lionelkusch (changes requested):

docs/src/glossary_and_notations.rst
jpaillard: What I intended is a section "minus index" with subsections groups and individual variables.

jpaillard: That makes sense, I added a crossref from the notaion to this entry in the Glossary.

docs/src/glossary_and_notations.rst
```diff
@@ -0,0 +1,66 @@
...
+    Permutation Feature Importance
+        The Permutation Feature Importance (PFI) is a measure of feature importance 
+        that consists in permuting the values of the feature of interest and measuring 
+        the performance drop triggered by this perturbation.
+    
+    :math:`X^{-j}`
+        See :term:`individual features`.
+    
+    :math:`X_{-G}`
+        See :term:`Feature groups`.
```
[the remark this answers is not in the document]
jpaillard: I see, done

lionelkusch:
> We'll have to add references to the glossary for CPI, PFI, LOCO etc. but this will come later when the documentation is a bit more advanced.

The glossary won't be the place for CFI, PFI, LOCO, ... because there will have their own documentation.
The principle of the glossary is to define the terms which are generic over all the implementations but don't have already a place to have been defined.

lionelkusch:
In this case, it will have a different purpose than the one proposed by scikitlearn.

The problem of having the definition in multiple places is to keep the consistency between all the definitions. Without tools dedicated to it, it will be a nightmare to maintain.

bthirion:
I see your point. But to me the first danger is jargon: giving users the impression that we use a cryptic language.

lionelkusch:
The jargon will come from the name of the methods and we already have the list of the API and the user guidelines which will define this term, I don't see the point here to add third place.

The only thing that we can add is the definition of word use in the name of methods such as knockoff, distillation, conditional tests, ...

jpaillard:
Done

bthirion (approved):

Issue #422, opened 2025-09-15
[DOC] add minimal examples


jpaillard:
We should have one "minimal textbook example" per method to show the very basic usage and add a visualization that can be used in the user guide. 
For the moment, let's make a bucket list of the methods that are part of the API v2 
- [ ] Conditional Feature Importance 
- [ ] LOCO
- [ ] D0CRT
- [ ] Permutation Feature Importance

lionelkusch:
Do you that the snippet of code in the PR #402 is enough or do you want something additional?

PR #439, opened 2025-09-17
[DOC] add minimal example for CFI

jpaillard:
Minimal example to complement the User guide.
The idea is to briefly show how to use the method, refer to the example in the user guide and maybe use the figure in the user guide. 
Related to #422

### DIFF
--- a/test/test_conditional_feature_importance.py
+++ b/test/test_conditional_feature_importance.py
@@ -571,6 +571,7 @@ class TestCFIExceptions:
         ):
             cfi.importance(X, y)
 
+
 @pytest.mark.parametrize(
     "n_samples, n_features, support_size, rho, seed, value, signal_noise_ratio, rho_serial",
     [(10, 10, 1, 0.2, 0, 1.0, 1.0, 0.0)],


src/hidimstat/visualization.py
jpaillard, line 1: I think it makes sense to have a plot_importance function since it will be used extensively in examples. 
However, I'm not sure where to put it. Would it make more sense to have a class method ` FeatureImportance.plot_importance()` ?

lionelkusch: The creation of a function for plotting adds an additional dependency on the plot, which is not essential.
Scikitlearn manages it by creating an extra class: https://scikit-learn.org/dev/developers/plotting.html

The other possibility is to add a submodule to it with function.

@bthirion what do you think?

lionelkusch: In this case, the best is to follow the guideline of scikit learn and add a condition for the import.

jpaillard: I think it's something like: 
```python 
try:
    import matplotlib.pyplot as plt
    import seaborn as sns
except ImportError:
    raise Exception("You need to install seaborn for using this functionality")
```

jpaillard:
I added the test for the plot using pytest-mpl. @Remi-Gau, do you have an opinion or suggestions on testing plotting functions?
It seems to be more sensitive to changes in the dependencies than other tests and thus more maintenance work. But maybe that's the price to pay to have plotting functions.

src/hidimstat/base_variable_importance.py
```diff
@@ -131,3 +132,71 @@ def _check_importance(self):
...
+        ax=None,
+        ascending=False,
+        **kwargs,
+    ):
+        """
+        Plot feature importances as a horizontal bar plot.
+
+        Parameters
+        ----------
+        importances : array-like of shape (n_features,) or (n_repeats, n_features)
```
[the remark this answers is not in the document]
jpaillard: removed

examples/plot_cfi_minimal.py
```diff
@@ -0,0 +1,96 @@
+"""
```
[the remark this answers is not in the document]
jpaillard: done

src/hidimstat/base_variable_importance.py
```diff
@@ -131,3 +132,71 @@ def _check_importance(self):
...
+            }
+        else:
+            df_plot = {
+                "Feature": feature_names,
+                "Importance": self.importances_,
+            }
+
+        df_plot = pd.DataFrame(df_plot)
+
+        # Sort features by decreasing mean importance
```
lionelkusch, line 189: Normally, the importance is only one value in the API.
The other case, if the model uses CV or bootstrapping, but in this case, the importance won't have multiple values, there will be another attribute for it.

test/test_conditional_feature_importance.py
```diff
@@ -569,3 +570,85 @@ def test_groups_warning(self, data_generator):
...
+        X, y, test_size=0.5, random_state=0
+    )
+    fitted_model = LinearRegression().fit(X_train, y_train)
+    cfi = CFI(
+        estimator=fitted_model,
+        imputation_model_continuous=LinearRegression(),
+        random_state=0,
+    )
+    # Make the plot independent of data / randomness to test only the plotting function
+    cfi.importances_ = np.stack(
```
lionelkusch, line 619: This is not realist cases, actually.

lionelkusch (commented):
We need to decide what is in `importances_` because in API 2, `importances_` is only an array of 1 dimension with a unique score for all the features.

test/test_conditional_feature_importance.py
jpaillard: What we want to test here is the behavior of the plot function, not the importance. So I keep the importance fixed. Changes in the behavior of the importance method (for instance, changes in the random generator) should not trigger this test.

jpaillard:
I let you handle which versions should run the mpl testing in #444 
I just specified the path where the reference images are located with `--mpl-baseline-path=tests/baseline_plots`

bthirion:
> > Having a single `importances_` attribute that could be either 1d (n_features) or 2d (n_folds, n_features) is more intuitive for the user. It is also compatible with scipy.stat's API for statistical tests. that we will use for selection. I think that this is where the FeatureImportanceCV class is relevant. I would change CFI to CFICV in the second test.
> 
> Cross validation and bootstrapping should be handled differently. I don't think that changing the dimension of importances_ is a good thing to expect because there are methods to aggregate them. It should be applied to other attributes, such as importances_cv_ .
> 
> What do you think?
> 
> Moreover, if we plan to support Multioutput estimators because in this case, how to make the difference between the multi output regression and the CV?

Could you remind us how you want to handle multiple values for importance: have we defined an attribute for that ?

jpaillard:
At the moment, we have an attribute importances_ of shape `n_features`. To measure importance with k-fold CV, you need to create k instances and then concatenate their importances. 
IMO, the best way would be to have an object `FeatureImportance` with `importances_ ` of shape `(n_samples_test, n_features)` and an object FeatureImportanceCV with `importances_ ` of shape `(n_folds, n_features)`. That way, you can directly pipe the importances_ into scipy.stat's API.

lionelkusch:
For the question of multioutput, see the issue #448.

bthirion:
> At the moment, we have an attribute importances_ of shape `n_features`. To measure importance with k-fold CV, you need to create k instances and then concatenate their importances. IMO, the best way would be to have an object `FeatureImportance` with `importances_ ` of shape `(n_samples_test, n_features)` and an object FeatureImportanceCV with `importances_ ` of shape `(n_folds, n_features)`. That way, you can directly pipe the importances_ into scipy.stat's API.

I don't really understand. Currently we have a `BaseVariableImportance` slass. How would `FeatureImportance` be related to that ?

examples/plot_cfi.py
```diff
@@ -0,0 +1,100 @@
...
+# ----------------------------------------
+# Finally, we visualize the importance of each feature using a bar plot.
+
+import matplotlib.pyplot as plt
+
+_, ax = plt.subplots(figsize=(6, 3))
+ax = cfi.plot_importance(ax=ax)
+ax.set_xlabel("Feature Importance")
+plt.tight_layout()
+
```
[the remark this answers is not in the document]
jpaillard: My knowledge about wine is quite limited 🙁 I could find on this page https://search.r-project.org/CRAN/refmans/ContaminatedMixt/html/wine.html that the 3 classes are Barolo, Grignolino, and Barbera. Would you have more insights? 

The only interpretation I could give is that, despite very large marginal importance, flavanoids and total_phenols are not picked by CFI, probably due to their high correlation (0.86 between these two) and their redundancy with other features.

src/hidimstat/base_variable_importance.py
jpaillard: Discussed in #449

test/test_conditional_feature_importance.py
jpaillard: Discussed in #449

PR #441, opened 2025-09-17
[DOC] add minimal example for LOCO

jpaillard:


examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,128 @@
+"""
+Leave-One-Covariate-Out (LOCO) feature importance with different regression models
+==================================================================================
+
+This example demonstrates how to compare LOCO feature importance across different
```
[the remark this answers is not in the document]
jpaillard: done

examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,128 @@
...
+# of features that exclude the feature of interest.
+
+X, y, beta, _ = multivariate_simulation(
+    n_samples=500,
+    n_features=10,
+    support_size=5,
+    rho=0.3,
+    signal_noise_ratio=8,
+    seed=0,
+)
```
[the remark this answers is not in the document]
jpaillard: Ok, I replaced it with sklearn's dataset

examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,143 @@
...
+
+This example demonstrates how to compare LOCO feature importance :footcite:t:`Williamson_General_2023` across different
+predictive models on the same regression dataset. LOCO is model-agnostic and can be
+applied to any predictive model. Here, we use a linear model, a random forest, a neural
+network, and a support vector machine. We compare the models based on their predictive
+performance (R2 score) and the LOCO feature importance they yield.
+"""
+
+# %%
+import pandas as pd
```
bthirion, line 13: I think it is not great to burden the reader with 8 lines of import. 
Let's do the import where they are needed.

examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,143 @@
...
+
+# %%
+# Loading and preparing the data
+# ------------------------------
+# We begin by simulating a regression dataset with 10 features, 5 of which
+# are in the support set, meaning they contribute to generating the outcome.
+# The data is then split into training and test sets. These sets are used both to fit
+# the predictive models and within the LOCO procedure, which refits models on subsets
+# of features that exclude the feature of interest.
+
```
bthirion, line 31: Wouldn't it be possible to use a public dataset instead ?
Because the data generation process is not really transparent for the reader.

examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,143 @@
...
+X, y, beta = make_regression(
+    n_samples=300,
+    n_features=10,
+    n_informative=5,
+    random_state=0,
+    coef=True,
+    noise=10.0,
+)
+beta = beta != 0
```
bthirion, line 40: Why this line ?

bthirion: Emphasize that this represents the ground truth.

examples/plot_loco_minimal.py
bthirion: OK, I see why: you want a ground truth. But please explain it !

examples/plot_loco_minimal.py
```diff
@@ -0,0 +1,143 @@
+"""
```
bthirion, line 1: I don't like this naming. Why not `plot_loco.py` ?
The example is not minimal bzw ;-)

bthirion (commented):
Thx, it looks pretty good.

examples/plot_loco_minimal.py
jpaillard: Yes, it's clearer this way.

bthirion (approved):
LGTM, thx !

examples/plot_loco.py
```diff
@@ -0,0 +1,149 @@
...
+    n_features=10,
+    n_informative=5,
+    random_state=0,
+    coef=True,
+    noise=10.0,
+)
+
+# We convert the coefficients of the data-generating process into a binary array
+# indicating the true support set of features.
+beta = beta != 0
```
lionelkusch, line 37:
```suggestion
beta = beta != 0

```

I think that it's better to have a separation.

examples/plot_loco.py
```diff
@@ -0,0 +1,149 @@
...
+    # contributions.
+    loco.fit(X_train, y_train)
+    importances = loco.importance(X_test, y_test)["importance"]
+    df_list.append(
+        pd.DataFrame(
+            {
+                "feature": list(range(X.shape[1])),
+                "importance": importances,
+                "model": model.__class__.__name__,
+                "support": beta,
```
lionelkusch, line 91: Why do you add beta?

It should be the same for all the models.

examples/plot_loco.py
```diff
@@ -0,0 +1,149 @@
...
+    # For each feature, remove it from the dataset, refit the model, and compute LOCO
+    # importance. This process is repeated for all features to assess their individual
+    # contributions.
+    loco.fit(X_train, y_train)
+    importances = loco.importance(X_test, y_test)["importance"]
+    df_list.append(
+        pd.DataFrame(
+            {
+                "feature": list(range(X.shape[1])),
+                "importance": importances,
```
lionelkusch, line 89: I propose to plot the relative importance of the score for having a better comparison between the different models. 
What do you think?

lionelkusch (commented):
I have few suggestions for improving this example.

examples/plot_loco.py
jpaillard: I don't understand, what do you mean?

lionelkusch (approved):
thank you for the example.

Issue #360, opened 2025-08-28
Control of random generator


lionelkusch:
For all the children classes of `BasePerturbation`, the randomness only appears at the `.predict()` step. So, for clarity, I suggest we only set/check `random_state` starting from that stage. 

In addition, it would be desirable that the behavior of the fitted class is deterministic: calling twice  `.importance()` on the same data gives the same result. 

I suggest that we expose the `random_state` at the importance/predict methods, but not before. 

For compatibility with CFI, this requires modifying the `ConditionalSampler`, which currently takes the `random_state` at __init__ but only uses it at `.sample()` we could thus make it an argument of `.sample(random_state=None)`

_Originally posted by @jpaillard in https://github.com/mind-inria/hidimstat/pull/344#discussion_r2307234877_

lionelkusch:
Agreed. However, exposing the random_state and setting its default value to None keeps this as a default behavior. But it allows, in addition, to control the randomness if desired. 

From the developer's perspective, I think it would clarify the code, narrowing the sections where randomness appears/is controlled.

_Originally posted by @jpaillard in https://github.com/mind-inria/hidimstat/pull/344#discussion_r2307282129_

jpaillard:
Sorry if I was not clear but [`GaussianMixture` does what I try to describe](https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7eacfaff061cf68e09bcd48aa93d4d6b/sklearn/mixture/_base.py#L437).

```python
class BaseMixture() ...
   ...
   def __init__(self, random_state=None, ...):
      self.random_state = random_state
      ...

   def sample(self, ...): 
      rng = check_random_state(self.random_state) 
...
```

In this way, if you let `random_state=None` multiple calls of sample lead to different random draws. But if you pass an int you get a deterministic behavior. 
You can try this snippet to check: 

``` python 
import numpy as np
from sklearn.mixture import GaussianMixture

for random_state in [0, None]:
    X = np.random.rand(100, 2)
    mixture = GaussianMixture(n_components=2, random_state=random_state)
    mixture.fit(X)

    samples_1 = mixture.sample(n_samples=10)
    samples_2 = mixture.sample(n_samples=10)

    eq_samples = np.array_equal(samples_1[0], samples_2[0])
    print(f"Random state: {random_state}, Equal samples: {eq_samples}")
```

It is then just a question of whether we want to expose the random_state in the `__init__` or `.predict()`, I am OK with both, 


[...]

PR #461, opened 2025-09-25
[DOC] add minimal developer guidelines in developer documentation

jpaillard:
I suggest beginning a Developer Guidelines section in Developer Documentation. 
For the moment, it just gathers and documents the feedback and decisions made in #360


lionelkusch (approved):
Thank you for the documentation.
I think for a start this is enough; we will complete later with some consensus on the different issues.

PR #474, opened 2025-10-03
[Tests] fix unstable tests for desparsified Lasso and knockoffs

jpaillard:
- Reduce the dimensionality of unstable tests
 - Change the condition for testing


lionelkusch:
Are you sure that you want to fix the seed for the test?

lionelkusch:
> Idk what the best practice is, I just thought that if we seed the simulation, we should also seed the function.

For the test, it's better to not add a seed because it allow to have a larger exploration of the usage of the code.
Normally, `pytest-randomly` set a global seed and allow to reproduce the result.

bthirion:
> Ok, I removed the random_state. That made me realize that even with the changes, tests remained unstable. Since the guarantees of these methods are in proba, I suggest going for a test of the FDP, and a test that the method has sufficient power rather than exact support recovery. So I suggest testing that the FDP is smaller than the chosen significance threshold for the desparsified LASSO and FDR for knockoffs. + a test that the power is greater than 0.8 (arbitrary).
> 
> This requires increasing the dimensionality and number of samples so that the proportions are more stable.

Isn't it possible to have en elementary test ? e.g. true beta within the confidence interval at least in 10 trials among 20.
otherwise, it's legitimate to seed the test in such a case.

test/test_desparsified_lasso.py
```diff
@@ -32,78 +38,92 @@ def test_desparsified_lasso():
...
+
+    correct_interval = np.sum((beta >= cb_min) & (beta <= cb_max))
+    assert correct_interval >= int(0.7 * n_features)
+
+    # Check p-values for important and non-important features
+    important = beta != 0
+    non_important = beta == 0
+    tp = np.sum(pval_corr[important] < alpha)
+    fp = np.sum(pval_corr[non_important] < alpha)
+    assert fp / np.sum(non_important) <= alpha
```
bthirion, line 57: Note that this holds only in expectation. The test could fail.

test/test_desparsified_lasso.py
```diff
@@ -32,78 +38,92 @@ def test_desparsified_lasso():
...
-    assert_equal(cb_max > beta, True)
-    assert_almost_equal(pval_corr, expected_pval_corr, decimal=1)
+    # Check that beta is within the confidence intervals
+    correct_interval = np.sum((beta >= cb_min) & (beta <= cb_max))
+    assert correct_interval >= int(0.7 * n_features)
+
+    # Check p-values for important and non-important features
+    tp = np.sum(pval_corr[important] < alpha)
+    fp = np.sum(pval_corr[non_important] < alpha)
+    assert fp / np.sum(non_important) <= alpha
```
bthirion, line 71: Same here.

test/test_desparsified_lasso.py
```diff
@@ -32,78 +38,92 @@ def test_desparsified_lasso():
...
-    pval, pval_corr, one_minus_pval, one_minus_pval_corr = (
-        desparsified_group_lasso_pvalue(beta_hat, theta_hat, precision_diag, test="F")
+    _, pval_corr, _, _ = desparsified_group_lasso_pvalue(
+        beta_hat, theta_hat, precision_diag
     )
 
     assert_almost_equal(beta_hat, beta, decimal=1)
-    assert_almost_equal(pval_corr, expected_pval_corr, decimal=1)
+    tp = np.sum(pval_corr[important] < alpha)
+    fp = np.sum(pval_corr[non_important] < alpha)
```
bthirion, line 124: here too

test/test_knockoff.py
```diff
@@ -205,38 +205,44 @@ def test_model_x_knockoff_exception():
...
         n_bootstraps=1,
-        random_state=seed,
         fdr=fdr,
     )
-    for i in selected:
-        assert np.any(i == non_zero)
-    selected, test_scores, threshold, X_tildes = model_x_knockoff(
+    tp = len(set(selected) & set(non_zero))
+    fp = len(set(selected) - set(non_zero))
+    assert fp / (p - len(non_zero)) <= fdr
```
bthirion, line 228: here too.

test/test_knockoff.py
```diff
@@ -205,38 +205,44 @@ def test_model_x_knockoff_exception():
...
         n_bootstraps=1,
-        random_state=seed,
         fdr=fdr,
     )
-    for i in selected:
-        assert np.any(i == non_zero)
+
+    tp = len(set(selected) & set(non_zero))
+    fp = len(set(selected) - set(non_zero))
+    assert fp / (p - len(non_zero)) <= fdr
```
bthirion, line 244: here too.

bthirion (approved):
LGTM, thx.

PR #491, opened 2025-10-15
[DOC] fix rendering of sampling classes

jpaillard:


docs/src/api.rst
```diff
@@ -13,9 +13,9 @@ Base Classes
    :toctree: ./generated/api/base/
    :template: class.rst
 
-   base_variable_importance.BaseVariableImportance
```
lionelkusch: I put the full name because there are not defined in the init at the root of the library.

You should play with the parameter of autosummary.
https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html

PR #516, opened 2025-10-27
[API 2] Add PDP 

jpaillard:
This is a second version of the effort initiated in #318. 
 - I tried to rely more heavily on sklearn's implementation. 
 - The idea is to have a first minimal tool for visualizing dependencies in the library


jpaillard:
Made an error when creating the branch

PR #517, opened 2025-10-27
[API 2] add PDP

jpaillard:
This work continues the efforts initiated in #318 
It relies more heavily on sklearn's implementation. I suggest having this as a first minimal visualization tool in the library.


bthirion (approved):
LGTM, thx !

PR #525, opened 2025-11-07
[MNT] clean-up source imports

jpaillard:
- Rename functions of the new API with the suffix "analysis" as introduced in #372. For Methods that have not yet been refactored, I will do it in their respective PRs
- Remove reid, quantile_aggregation from the source imports

### DIFF
--- a/docs/src/api.rst
+++ b/docs/src/api.rst
@@ -38,16 +38,16 @@ Feature Importance functions
    :toctree: ./generated/api/class/
    :template: function.rst
    
-   cfi_analysis
+   cfi_importance
    clustered_inference
    clustered_inference_pvalue
-   d0crt_analysis
-   desparsified_lasso_analysis
-   model_x_knockoff_analysis
+   d0crt_importance
+   desparsified_lasso_importance
+   model_x_knockoff_importance
    ensemble_clustered_inference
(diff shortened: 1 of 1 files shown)


bthirion:
The PR is technically correct. I'm not a big fan of appending the "_analysis" which makes name longer without adding much. Bit IIRC you wanted that for the sake of clarity ?

jpaillard:
Yes, maybe we can find a better alternative.  
I think having loco and LOCO (same for all other methods) that only differ from using capital letters is error prone.  
In sklearn for instance they have Ridge and ridge_regression, which I belive makes it less likely for a user to make mistakes in imports.

bthirion:
> Yes, maybe we can find a better alternative. I think having loco and LOCO (same for all other methods) that only differ from using capital letters is error prone. In sklearn for instance they have Ridge and ridge_regression, which I belive makes it less likely for a user to make mistakes in imports.

OK, so then, let's  take loco_importance for the time being.

bthirion (approved):

jpaillard:
@bthirion, I renamed all functions with the suffix API. 
There's also a problem with the rendering of their docstring, which was not introduced in this PR. I reported it in an issue: #528

PR #530, opened 2025-11-10
[API 2] Add Cross validation for perturbation based methods

jpaillard:
- This PR builds on top of #449, but the diff became too large, so it became easier to start over from main. 
- It implements  `CFImportanceCV`, `PFImportanceCV`, and `LOCOImportanceCV`. I am open to suggestions for the name
- I also took this opportunity to rewrite the diabetes example.


src/hidimstat/conditional_feature_importance.py
```diff
@@ -289,3 +289,111 @@ def cfi_importance(
         P-values for importance scores. 
     """,
 )
+
+
+class CFImportanceCV(BasePerturbationCV):
```
bthirion: It would be more consistent to call it 'CFICV' ?

test/test_permutation_feature_importance.py
```diff
@@ -229,3 +229,48 @@ def test_pfi_reproducibility_with_rng(pfi_test_data):
...
+)
+def test_pfi_cv(data_generator):
+    """
+    Test that PFI with cross-validated estimator works as expected. In particular,
+        - Empirical FDP is below the target FDR level
+        - Power is above 0.8, which is an arbitrary threshold
+
+    Notes
+    -----
+     - Even though the only the expected FDP should be controlled, in practice
```
bthirion:
```suggestion
     -  Although only the expected FDP is controlled in theory, in practice
```

test/test_conditional_feature_importance.py
```diff
@@ -848,3 +850,43 @@ def test_cfi_reproducibility_with_rng(cfi_test_data):
...
+        estimators=estimators,
+        cv=cv,
+        random_state=0,
+        n_jobs=5,
+    )
+    cfi_cv.fit_importance(X, y)
+    selected = cfi_cv.fdr_selection(fdr=0.05)
+
+    alpha = 0.05
+    tp = np.sum([int(i) in important_features for i in np.where(selected)[0]])
```
bthirion: this is a bit awkward, but could be addressed in future PR

bthirion (approved):
LGTM, thx !

test/test_conditional_feature_importance.py
jpaillard: I changed it to use the fdp_power function instead.

PR #533, opened 2025-11-16
[API 2] CluDL and EnCluDL

jpaillard:


examples/plot_2D_simulation_example.py
```diff
@@ -1,361 +1,322 @@
...
-significantly the spatial uncertainty compared to clustered inference
-algorithms which consider only one spatial compression.
-
-References
-----------
-.. footbibliography::
-
+In this example, we present how to perform inference on 2D simulated data. This setting
+is particularly challenging since the number of features (pixels in 2D) is much larger
+than the number of samples. We first illustrates the limitations of the Desparsified
```
bthirion:
```suggestion
than the number of samples. We first illustrate the limitations of the Desparsified
```

examples/plot_2D_simulation_example.py
```diff
@@ -1,361 +1,322 @@
...
+ax.legend(handles=legend_handles, loc="lower center")
+plt.tight_layout()
 
-# ensemble of clustered desparsified lasso (EnCluDL)
-list_ward, list_desparsified_lasso = ensemble_clustered_inference(
-    X_init,
-    y,
-    ward,
-    scaler_sampling=StandardScaler(),
+
```
bthirion: Maybe add a short comment here

bthirion (commented):
Started a pass on simulated data example.

examples/plot_2D_simulation_example.py
jpaillard: done

test/test_base_variable_importance.py
```diff
@@ -168,18 +169,16 @@ def test_selection_fdr_one_minus_pvalue(self, set_100_variable_sorted):
         vi = set_100_variable_sorted
         true_value = np.arange(100) >= 34
         selection = vi.fdr_selection(fdr=0.9, alternative_hypothesis=True)
```
bthirion: This test does simply not make sense. It can be removed.

bthirion (commented):
LGTM overall.

examples/plot_fmri_data_example.py
lionelkusch: If you remove the try except, you should remove these lines.

src/hidimstat/ensemble_clustered_inference.py
```diff
@@ -1,593 +1,446 @@
...
+        Returns
+        -------
+        self : EnCluDL
+            Fitted estimator.
+        """
+        rng = check_random_state(self.random_state)
+
+        def _fit_one(rng_spawned):
+            clu_dl = CluDL(
+                desparsified_lasso=deepcopy(self.desparsified_lasso),
```
lionelkusch: You should not use deepcopy but clone from sciklearn because deepcopy can't have some issue.

examples/plot_digits.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+# %%
+# Performing inference on handwritten digits is challenging because the images are not
+# perfectly aligned, and the pixel regions occupied by digits vary from one image to
+# another. However, using EnCluDL, we can still identify clusters of pixels that are
+# statistically significant for the classification tasks. For example, we can detect
+# the bottom left portion of the loop when distinguishing between digits 0 and 9, the
+# bottom left corner of digit 4 when distinguishing between digits 4 and 7, and the
+# central vertical stroke when distinguishing between digits 0 and 1.
```
bthirion: Just wondering: does the result look nicer with more bootstraps ? If yes, it would be worth doing it.

test/test_ensemble_clustered_inference.py
```diff
@@ -6,92 +6,165 @@
...
+from hidimstat._utils.scenario import (
+    multivariate_simulation,
+    multivariate_simulation_spatial,
 )
+from hidimstat.statistical_tools.multiple_testing import fdp_power
 
 
-# Scenario 1: data with no temporal dimension
-def test_clustered_inference_no_temporal():
+def fdp_power_spatially_relaxed(
```
bthirion:
```suggestion
def spatially_relaxed_fdp_power(
```

bthirion (commented):
LGTM.

examples/plot_digits.py
jpaillard: Unfortunatly no. I tried with 50 and it's almost the same. WDYT about this example? Should it be merged?

bthirion (approved):

PR #541, opened 2025-12-01
[CORE] remove `one_minus_pval` attribute, add `fwer_selection` 

jpaillard:
### Changes in the core API 
 - Remove the attribute `one_minus_pval` from `BaseVariableImportance`, it is now only computed as an intermediate step of `fdr_selection` when `two_tailed_test=True`
 - `one_minus_pval` was used in some examples for FWER control. I also added a method `fwer_selection` which currently only supports the Bonferroni procedure.  
 - The output of both methods is a mask with 0 if non-selected, 1 if selected. If `two_tailed_test=True`, -1 for negative effects. 
 - By default, for `BaseVariableImportance`, `two_tailed_test=False`. For DL, CluDL, EnCluDL, this default is overwritten to `True`

### Other changes (most of the diff) 
 - update examples using `one_minus_pval`. Replace with `fwer_selection`
 - Add tests for coverage of new methods
 - Modify existing tests. To test fdr_selection, repeat multiple draws and test that the empirical fdr is below the target.

### DIFF
--- a/test/test_base_variable_importance.py
+++ b/test/test_base_variable_importance.py
@@ -489,6 +489,9 @@ def test_fwer_selection():
     assert np.mean(false_discovery_list) < target_fdr + test_tol
     assert np.mean(power_list) > 0.4 - test_tol
 
+    with pytest.raises(ValueError, match="Only 'bonferroni' procedure is supported"):
+        vim.fwer_selection(fwer=0.1, procedure="invalid_procedure")
+
 
 def test_clustered_fwer_selection():
     """


examples/plot_fmri_data_example.py
```diff
@@ -268,15 +270,19 @@ def plot_map(
     )
 
 
+pval_cludl, _, one_minus_pval_cludl, _ = pval_from_two_sided_pval_and_sign(
```
bthirion: Can't we avoid the `one_minus_pval_cludl` here ?

src/hidimstat/base_variable_importance.py
```diff
@@ -272,39 +280,90 @@ def fdr_selection(
...
             )
-            selected_pvalues = self.pvalues_ <= threshold_pvalues
-        else:
-            selected_pvalues = np.zeros_like(self.pvalues_, dtype=bool)
-
-        # selection on 1-pvalue
-        if alternative_hypothesis is None or alternative_hypothesis:
             threshold_one_minus_pvalues = fdr_threshold(
-                self.one_minus_pvalues_,
+                one_minus_pval,
```
bthirion: can't we get rid of it ?

src/hidimstat/base_variable_importance.py
```diff
@@ -272,39 +280,90 @@ def fdr_selection(
...
+        Performs feature selection based on Family-Wise Error Rate (FWER) control.
+
+        Parameters
+        ----------
+        fwer : float
+            The target family-wise error rate level (between 0 and 1)
+        procedure : {'bonferroni'}, default='bonferroni'
+            The FWER control method to use:
+            - 'bonferroni': Bonferroni correction
+        n_tests : int or None, default=None
```
bthirion: why would `n_tests` differ from `len(self.pvalues_)` ?

src/hidimstat/base_variable_importance.py
```diff
@@ -272,39 +280,90 @@ def fdr_selection(
...
+            if n_tests is None:
+                if hasattr(self, "clustering_"):
+                    print("Using number of clusters for multiple testing correction.")
+                    n_tests = self.clustering_.n_clusters_
+                else:
+                    print("Using number of features for multiple testing correction.")
+                    n_tests = self.importances_.shape[0]
+            threshold_pvalue = fwer / n_tests
+            selected = (self.pvalues_ < threshold_pvalue).astype(int)
+            if two_tailed_test:
```
bthirion: here we should set threshold_pvalue to fwer / (2 * n_tests)

test/test_base_variable_importance.py
```diff
@@ -113,47 +117,56 @@ def test_selection_threshold_max(self, set_100_variable_sorted):
...
+    """
+    bh_fdp_list = []
+    bhy_fdp_list = []
+    bh_power_list = []
+    bhy_power_list = []
+    n_features = 100
+    target_fdr = 0.1
+    test_tol = 0.1
+
+    for _ in range(500):
```
bthirion: oops, that's a lot !

src/hidimstat/base_variable_importance.py
jpaillard: done

test/test_desparsified_lasso.py
```diff
@@ -93,82 +109,96 @@ def test_desparsified_group_lasso():
...
+            ).fit(X, y)
+        importances = desparsified_lasso.importance()
+
+        assert_almost_equal(importances, beta, decimal=1)
+
+        important = beta[:, 0] != 0
+
+        assert_almost_equal(importances, beta, decimal=1)
+        selected = desparsified_lasso.fwer_selection(fwer=alpha, two_tailed_test=False)
+        fdp, power = fdp_power(np.where(selected)[0], np.where(important)[0])
```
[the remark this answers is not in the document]
jpaillard: I replaced by a test that the empirical FWER (over 10 repeats) is lower than the target.

examples/plot_2D_simulation_example.py
```diff
@@ -86,9 +86,7 @@
 dl.fit_importance(X_init, y)
 
 # compute estimated support (first method)
-selected_dl = (dl.pvalues_ < (fwer_target / 2) / n_features) | (
-    dl.one_minus_pvalues_ < (fwer_target / 2) / n_features
-)
+selected_dl = dl.fwer_selection(fwer=fwer_target / 2, n_tests=n_features)
```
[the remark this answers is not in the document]
bthirion: can we fix that ?

src/hidimstat/base_variable_importance.py
bthirion: Thx. Can you remove the division by 2 from examples ?

examples/plot_2D_simulation_example.py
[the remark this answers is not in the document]
jpaillard: yes, I forgot to push the changes on the 2D simulation when making the reviews. It should be good now.

bthirion:
One line remains untested. Cna you make sure to test it ? Thx !

bthirion (approved):
LGTM, thx !

PR #542, opened 2025-12-04
[CORE] change fdp_power

jpaillard:


src/hidimstat/statistical_tools/multiple_testing.py
```diff
@@ -7,10 +7,11 @@ def fdp_power(selected, ground_truth):
 
     Parameters
     ----------
-    selected : ndarray
-        Array of indices of selected variables (R-style indexing)
-    ground_truth : ndarray
-        Array of true non-null variable indices
+    selected : ndarray (n_features,)
+        Array of selected variable. 0 for non-selected, 1 for selected, -1 for
+        selected with negative effect (for two-tailed tests)
```
bthirion: We don't really care about the sign here ? Maybe simply "zero" and "non-zero" ?

jpaillard: Yes, I can fix the docstring accordingly

bthirion (approved):
LGTM, thx !

PR #553, opened 2025-12-08
Permutation Feature Importance DOC

AngelReyero:


jpaillard:
It looks almost good to me. I still have comments regarding the `.bib` file, from which I suggest removing links and abstracts. 
Also, suggest adding at the end of the page of PFI, which will add links to the examples using PFI. 
```
Examples
--------

.. minigallery:: hidimstat.PFI
```

I see that the test are not passing but merging #558 shoud improve

bthirion:
As you wish. But this PR is already in pretty good shape, I'm not against merging it.

PR #569, opened 2026-01-08
568 test timeout issue in minimal version test

jpaillard:


bthirion:
> This should effectively increase the timeout to 5min for the tests with minimal version of the dependencies only. See the CI passing in [05d66eb](https://github.com/mind-inria/hidimstat/pull/569/commits/05d66eb768f5636845a92fc2d66792a9c1194668) The [testCI] flag triggered the `ci_test.yml` action where I placed the modification I suggest. I then cleaned up the content of `ci_test.yml`.
> 
> Is it a suitable solution @bthirion?

I would try to keep it to 2min ?

test/test_ensemble_clustered_inference.py
```diff
@@ -172,20 +172,20 @@ def test_encludl_spatial():
...
             roi_size=roi_size,
-            spatial_tolerance=3,
+            spatial_tolerance=2,
             shape=shape,
         )
         fdp_list.append(fdp)
         power_list.append(power)
     assert np.mean(power_list) >= 0.5
-    assert np.mean(fdp_list) <= fdr
+    assert np.mean(fdp_list) <= fwer
```
bthirion: It's weird. fdps should be compared to an fdr level, not an fwer level.

Issue #570, opened 2026-01-27
improve scikit-learn compatibility


jpaillard:
Cloning hidimstat methods (such as CFI, LOCO,...) is currently not possible. 

```python 
import numpy as np
from sklearn.base import clone
from sklearn.linear_model import LinearRegression

from hidimstat import LOCO

X = np.random.randn(20, 2)
y = np.random.randn(20)
predictor = LinearRegression()
predictor.fit(X, y)
loco = LOCO(estimator=predictor)

loco_clone = clone(loco)
```
returns 
``` 
>>> NotFittedError: This LinearRegression instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator.
```

This is because we currently require the estimator passed to any `BasePerturbation` (LOCO, PFI, CFI ...) to be fitted (`check_is_fitted(estimator)` in the `__init__()` ). 

IMO, it makes sense to support fitted estimators (e.g., DL models trained separately), but we should also allow passing non-fitted estimators, which we would then fit in `BasePerturbation.fit()`. 
That would allow: 1. support unfitted estimators and 2. improve the compatibility with sklearn by allowing to clone hidimstat methods.

Issue #350, opened 2025-08-26
Run sklearn check_estimators on the hidimstat estimators that are expected to behave like a sklearn estimator


Remi-Gau:
See https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html

This could help standardize a few things in the API and behavior of the estimator.

I can open a PR to set up the tests for this to get you started if you are interested.

Remi-Gau:
I can also help navigate those different sklearn checks as I have now been playing with them for almost... (check notes)... almost a year...

lionelkusch:
It's a good idea, but in my opinion, it's a bit early to implement it because there is a need to be still flexible with the API because we need to have some feedback from users or some experience with the API before stiffening it.

Remi-Gau:
from personal experience: you want to have those checks running earlier rather than later because fixing things later will take A LOT OF TIME.

also the constraints on the API are not crazy strict: for a lot of them there are more of the type: cannot call score() on an unfitted estimator

also also: those checks are made in a way that allow gradual or optional opt-in: you can skip some checks if they don't make sense in your case.
For example most nilearn etimators have to skip the checks that work on numpy arrays only because nilearn estimators work on images.
Or you can skip or xfail some checks and just say "TODO: compliance with this check has not been implemented yet".

bthirion:
Thx for raising this, but indeed I think it would be a bit premature now.

PR #571, opened 2026-01-29
[MAINT] run sklearn checks on BasePerturbation and BasePerturbationCV

Remi-Gau:
- relates to #570 
- relates to #350

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
-=====================================
-Version 0.0.4 (under devellopement)
-=====================================
+===================================
+Version 0.4.0 (under devellopement)
+===================================
 
 HIGHLIGHTS
 ----------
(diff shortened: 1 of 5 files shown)


.github/workflows/call_linter.yml
```diff
@@ -8,6 +8,10 @@ on:
 # disable all the permission for the workflow
 permissions: {}
 
+# Force to use color
+env:
+    FORCE_COLOR: true
```
jpaillard: What does this do?

jpaillard (commented):
So IIUC, all the failed estimator checks are those we should include in the `expected_failed_checks` if we have a good reason for it?

test/test_ensemble_clustered_inference.py
Remi-Gau: unused import: 

ruff check --select F401 examples --fix

test/test_knockoff.py
```diff
@@ -14,6 +14,18 @@
 from hidimstat.samplers import GaussianKnockoffs
 from hidimstat.statistical_tools.multiple_testing import fdp_power
 
+# TODO
```
bthirion: Maybe open the issues corresponding to current failures ?

bthirion (commented):
Looks good overall.

PR #576, opened 2026-02-03
[FIX] improve compatibility BasePerturbation wth sklearn

Remi-Gau:
- relates to #350

### DIFF
--- a/src/hidimstat/base_perturbation.py
+++ b/src/hidimstat/base_perturbation.py
@@ -23,7 +23,8 @@ class BasePerturbation(BaseVariableImportance, GroupVariableImportanceMixin):
     Parameters
     ----------
     estimator : sklearn-compatible estimator
-        The estimator that will be used for predictions. It will be automatically fitted if it has not already been.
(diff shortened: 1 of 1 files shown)


src/hidimstat/base_variable_importance.py
```diff
@@ -501,32 +507,38 @@ def fit(self, X, y=None):
...
+                    for ids in list(self.features_groups_.values())
                 ]
         else:
             raise ValueError("features_groups needs to be a dictionary")
         return self
 
+    def __sklearn_is_fitted__(self) -> bool:
+        return (
+            getattr(self, "n_features_groups_", None) is None
+            or getattr(self, "_features_groups_ids", None) is None
```
jpaillard:
```suggestion
            getattr(self, "n_features_groups_", None) is not None
            or getattr(self, "_features_groups_ids", None) is not None
```
IIUC, this replaces L540, which was raising an error if either attribute was None. If the class is fitted, they should both **not** be None

PR #581, opened 2026-02-05
[MAINT] improve sklearn compatibility of BasePerturbationCV

Remi-Gau:
- relates to #350

### DIFF
--- a/docs/tools/conf.py
+++ b/docs/tools/conf.py
@@ -220,8 +220,8 @@ linkcheck_ignore = [
     r"https://github.com/*",
 ]
 
-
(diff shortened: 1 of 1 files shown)


src/hidimstat/base_perturbation.py
```diff
@@ -359,8 +359,8 @@ class BasePerturbationCV(BaseVariableImportance):
 
     def __init__(
         self,
-        estimators,
-        cv,
+        estimators=None,
+        cv=None,
```
Remi-Gau: estimators should have initializable with some defaults

Issue #574, opened 2026-02-03
a lot of errors that need a manual fix are ignored for now


jpaillard:
a lot of errors that need a manual fix are ignored for now

_Originally posted by @Remi-Gau in https://github.com/mind-inria/hidimstat/pull/573#discussion_r2757793200_

PR #593, opened 2026-02-09
[MAINT] use numpy.random.Generator as fixture in tests

Remi-Gau:
- fixes https://docs.astral.sh/ruff/rules/numpy-legacy-random/
- relates to #574


Remi-Gau:
> Just wondering: is there a computational advantage to it, or is it readability or "good practice" ?

from my reading of [this](https://docs.astral.sh/ruff/rules/numpy-legacy-random/#why-is-this-bad) and the other links in the page, I would put this more under 'good practice'

In this case it also allows for the introduction of a rng fixture that can help remove duplicate code across tests

lionelkusch:
@Remi-Gau  
I don't think that it's a good idea to fix the seed. I preferred to have random runs and a control through pytest-randomly than a fix seed, which create reduce the coverage of cases from the tests.

PR #599, opened 2026-02-11
[MAINT] Run sklearn check_estimators on CFI and CFICV

Remi-Gau:
- relates to #350


test/conftest.py
```diff
@@ -32,6 +33,13 @@ def rng():
     return _rng()
 
 
+def fitted_linear_regression():
```
bthirion: Can you add a one-line docstring ?

bthirion (commented):
LGTM overall.

test/conftest.py
```diff
@@ -34,7 +34,10 @@ def rng():
 
 
 def fitted_linear_regression():
+<<<<<<< sklearn/CFI
+=======
     """Return a fitted linear regression model."""
+>>>>>>> main
```
jpaillard:
```suggestion
```
Seems that these were added by mistake, no?

Remi-Gau:
hmmm... I think we have a CI setup issue here: PRs opened from a fork cannot seem to be able to run PRs
@jpaillard can you confirm?

jpaillard:
It looks like it. This was not a problem in the past, however: see https://github.com/mind-inria/hidimstat/pull/590

Remi-Gau:
I think that's related to the latest version of the checkout action that was just updated

https://github.com/actions/checkout#whats-new

bthirion (approved):
LGTM, thx.

Remi-Gau:
ok tests failures seem to only affect older dependencies (I suspect older sklearn)

.github/workflows/call_test_package.yml
```diff
@@ -24,7 +24,7 @@ jobs:
   build-test:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true # if on job fail all the other jobs are cancelled
+      fail-fast: false # if on job fail all the other jobs are cancelled
```
Remi-Gau, line 27:
```suggestion
      fail-fast: true # if on job fail all the other jobs are cancelled
```
revert before merge

PR #600, opened 2026-02-11
[FIX] inconsistent naming in `ModelXKnockoff`

### LINKED ISSUE #575: [BUG] naming problem in the `ModelXKnockoff`
The issue comes from a naming problem in the `ModelXKnockoff` class. We document the argument `ko_generator`, but create an attribute named `generator`. 
Similarly, `random_state` is set with a typo `randoms_sate`

_Originally posted by @jpaillard in https://github.com/mind-inria/hidimstat/pull/571#discussion_r2753091099_

jpaillard:
closes #575

### DIFF
--- a/.github/workflows/call_publish_result.yml
+++ b/.github/workflows/call_publish_result.yml
@@ -96,7 +96,7 @@ jobs:
     # name for files
     env:
       OS: 'macos-latest'
-      PYTHON: '3.13'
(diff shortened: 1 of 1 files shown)


jpaillard:
I added the test to check sklearn estimators. I think most of the checks listed in `expected_failed_checks` should be fixed. I will open a follow-up PR

PR #607, opened 2026-02-19
[DOC] add mne MEG example

jpaillard:


examples/plot_meg_somato.py
```diff
@@ -0,0 +1,226 @@
...
+mne.datasets.fetch_fsaverage(subjects_dir=subjects_dir)
+
+morph = mne.compute_source_morph(
+    src=forward["src"],
+    subject_from=f"{subject}",
+    subject_to="fsaverage",
+    subjects_dir=subjects_dir,
+    spacing=5,
+)
+stc_fsaverage = morph.apply(stc_pvals)
```
jpaillard: MNE 3D plotting requires additional dependencies, like PyVistaQt. To avoid this I tried to use nilearn's plotting functions.  
But mapping the identified cortical sources to a plottable stat_map is not so easy. The workaround I found is to use mne morphing, which requires downloading `mne.datasets.fetch_fsaverage(subjects_dir=subjects_dir)`. It's not a perfectly smooth solution. I would be open to alternative suggestions.

PR #632, opened 2026-04-13
[BUG] unexpected group order change in VariableImportance

### LINKED ISSUE #531: [BUG] `GroupVariableImportanceMixin` should not use `sorted(self.features_groups.keys())`
A bug was introduced in the class: [`GroupVariableImportanceMixin`](https://github.com/mind-inria/hidimstat/blame/c458319383866e4c93a5f1db0cf17a61a3e8426c/src/hidimstat/base_variable_importance.py#L443). 
The private attribute `_features_groups_ids ` is generated by iterating through: `sorted(self.features_groups.keys())`. This creates a discrepancy between the order of groups in `features_groups` (which is the order expected by the user) and `_features_groups_ids `, which is used to create the $X^{-j}, X^j$ sets and then the importance array. This is illustrated below; the user expects 'dummy group' to be the second group, but its importance will actually be located at the first index of the importance array. 

```python
import numpy as np
import pandas as pd
from sklearn.linear_model import LinearRegression

from hidimstat import CFI

X = pd.DataFrame(np.random.rand(100, 3), columns=["a", "b", "c"])
y = np.random.randn(100)

features_groups = {
    "group 1": ["a", "b"],
    "dummy group": ["c"],
}

estimator = LinearRegression()
estimator.fit(X, y)
cfi = CFI(estimator=estimator, features_groups=features_groups)
cfi.fit(X, y)
print(cfi.features_groups)
print(cfi._features_groups_ids) 
```
out:
```python
>>> features_groups: {'group 1': ['a', 'b'], 'dummy group': ['c']}
>>> _features_groups_ids [[2], [0, 1]]
```

jpaillard:
`test_feature_groups_order_preserved` illustrates the unexpected behavior. The test fails with the current implementation (see first commit) and should pass after the proposed fix.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Bug fixes
 ---------
 
 * Fixed a bug in CLuDL/EncluDL: cloning the internal estimators to avoid them carrying unwanted history. (PR #425)
+* Fix the unexpected change in the order of feature groups when a dataframe is used as input (PR #632)
 
 Maintenance
 -----------
--- a/src/hidimstat/base_variable_importance.py
+++ b/src/hidimstat/base_variable_importance.py
@@ -541,14 +541,14 @@ class GroupVariableImportanceMixin:
                 j: [j] for j in range(self.n_features_groups_)
             }
             self._features_groups_ids = np.array(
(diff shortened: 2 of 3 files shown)


bthirion (approved):
LGTM, thx

PR #643, opened 2026-04-17
[BUG] fix KO `threshold_mesh`

### LINKED ISSUE #639: [BUG] knockoff selection is wrong
https://github.com/mind-inria/hidimstat/blob/main/src/hidimstat/knockoffs.py#L491 should be
```
threshold_mesh =  np.concatenate([[0], threshold_mesh, [np.inf]])
```
Currently the line does nothing and the result is wrong (there is always a detection as long as there is nonzero positive coefficient !)

jpaillard:


bthirion:
Thx. You want to add a test for that one. I can provide it ifneedbe.

jpaillard:
If you have an example that used to fail and is now solved, that would be perfect.

bthirion:
I've pushed a test.

test/test_knockoff.py
```diff
@@ -267,6 +267,27 @@ def test_knockoff_function_not_centered():
...
+        estimator=GridSearchCV(
+            Lasso(), param_grid={"alpha": np.linspace(0.2, 0.3, 5)}
+        ),
+        preconfigure_lasso_path=None,
+    ).fit(X, y)
+    model_x_knockoff.importance()
+    selected = model_x_knockoff.fdr_selection(fdr=fdr)
+    fdp_power(selected, beta)
+    print(len(np.where(beta)[0]))
+    assert (not selected).all()
```
bthirion:
```suggestion
    assert ((selected == False).all())
```

bthirion (commented):
LGTM. Just need to fix the CI

test/test_knockoff.py
```diff
@@ -267,6 +267,27 @@ def test_knockoff_function_not_centered():
...
+        estimator=GridSearchCV(
+            Lasso(), param_grid={"alpha": np.linspace(0.2, 0.3, 5)}
+        ),
+        preconfigure_lasso_path=None,
+    ).fit(X, y)
+    model_x_knockoff.importance()
+    selected = model_x_knockoff.fdr_selection(fdr=fdr)
+    fdp_power(selected, beta)
+    print(len(np.where(beta)[0]))
+    assert ((selected == False).all())
```
jpaillard:
```suggestion
    assert not selected.any()
```
It seems that Ruff-linter is unhappy with this syntax. This should fix.

bthirion (approved):

PR #648, opened 2026-05-12
[DOC] Inlcuded naming conventions in dev guidelines

### LINKED ISSUE #39: Name of files
There is a nonhomogenous way to name files and the methods. 

We need to decide if we used a unique acronym for the names of files or/and functions or the full name.

For example, the file adaptive_permutation_threshold.py contains the function ada_svr, and the file loco.py contains loco.

Moreover, the test for knockoffs.py is the only one that has the name of the function test_model_x_knockoff.

GrituX:
Following #39, included naming conventions in developer guidelines: upper camel case for classes, and snake case for files, both without acronyms.

### DIFF
--- a/tools/documentation_developer/developer_guidelines.rst
+++ b/tools/documentation_developer/developer_guidelines.rst
@@ -67,4 +67,8 @@ Naming
 ~~~~~~
 Files should follow the `snake case <https://en.wikipedia.org/wiki/Snake_case>` convention,
 and classes should follow the `upper camel case <https://en.wikipedia.org/wiki/Camel_case>`
-convention. Names should not contain acronyms and should instead include the full version.
+convention. Names should not contain acronyms and should instead include the full version,
+except for really long names, such as Distilled Conditional Randomization Test
+abbreviated as D0CRT, or names that are standard in the scientific community, such as 
+LOCO which stands for Leave One Covariate Out. Functions should be written in snake case,
+and should not contain acronyms similarly to classes.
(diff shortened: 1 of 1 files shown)


tools/documentation_developer/developer_guidelines.rst
```diff
@@ -59,3 +59,12 @@ test patterns help to ensure consistent behavior:
...
 * ``test_<method>_reproducibility_with_rng``
+
+Writing Conventions
+-------------------
+
+Naming
+~~~~~~
+Files should follow the `snake case <https://en.wikipedia.org/wiki/Snake_case>` convention,
+and classes should follow the `upper camel case <https://en.wikipedia.org/wiki/Camel_case>`
+convention. Names should not contain acronyms and should instead include the full version.
```
bthirion: + functions are written in snake case

jpaillard: Should we allow some tolerance for the acronyms? Some names are really long otherwise (e.g., D0CRT Distilled Conditional Randomization Test). This kind of exception is also allowed in libraries like sklearn for Lasso, SVM ...

GrituX: Right, I forgot to add that.

GrituX: I think it would be nice to allow that indeed, as discussed in the related issue. We can add the exception for really long names, or for acronyms that are fairly standard in the community like LOCO ?

bthirion: Agreed. In general it's good to avoid acronyms but indeed this is not an absolute rule.

jpaillard (approved):

PR #649, opened 2026-05-19
[MAINT] D0CRT: remove extra term in variance of X-residual

### LINKED ISSUE #118: Difference between original code of DCRTO and the implementation in hidimstat
The implementation between the original repository of dcrt0 and the actual implementation of dcrt0 of hidimstat has significant differences in the computation of the variance of residuals of the distillation of x and the ts.

Difference 1:
https://github.com/mind-inria/hidimstat/blob/e710ea979d04d5b90e9e1d98b068d52261c4bf30/src/hidimstat/dcrt.py#L263-L265

[link](https://github.com/moleibobliu/Distillation-CRT/blob/dddc8cc4027c236233cb9f0afb65290fb622be15/functions.R#L642-L647)
```
    sigma2_x <- mean(x_res^2)
```
Difference 2:
https://github.com/mind-inria/hidimstat/blob/e710ea979d04d5b90e9e1d98b068d52261c4bf30/src/hidimstat/dcrt.py#L331
https://github.com/mind-inria/hidimstat/blob/e710ea979d04d5b90e9e1d98b068d52261c4bf30/src/hidimstat/dcrt.py#L228

[link](https://github.com/moleibobliu/Distillation-CRT/blob/dddc8cc4027c236233cb9f0afb65290fb622be15/functions.R#L642-L647)
```
imp_obe <- abs(mean(X_res * eps_res))
emp_var <- mean(eps_res^2) * sigma2_x
pvl <- 2 * pnorm(- sqrt(n) * abs(imp_obe) / sqrt(emp_var))
```

Do you have some justification of this difference? @bthirion @jpaillard @AngelReyero

jpaillard:
## Problem
In the current implementation of D0CRT, the term ` alpha * np.linalg.norm(model_x.coef_, ord=1)` is added to the estimate of the variance of the X-residual (part of each feature that is not explained by the others)  https://github.com/mind-inria/hidimstat/blob/d24810b9fa2dd0332155e51a87a74a909c13feb8/src/hidimstat/distilled_conditional_randomization_test.py#L765
No one really knows why this term is added (see #118). The fact that an L1 norm is used seems related to the fact that the default model-x is a Lasso, but idk. 

## This PR
I added an example `plot_D0CRT_comparison.py` which compares the current implementation with a patched version, removing this term.  It shows that this term makes the inference more conservative, reducing the power and the type-1 error rate slightly. 

Since the changes in behavior are minor, I would suggest removing the additional term to be closer to the original description in the paper.


jpaillard:
The idea would be that you take a look at the example and, if you agree, I will remove only the line discussed and add a regression test to show what changed and what did not.

bthirion:
Isn't it risky if it makes the estimator anti-conservative ?

bthirion:
I would say that the trick came from a paper but if we can't trace it that's annoying indeed.

jpaillard:
In theory, it should still control the type-1 error rate, and it is the implementation that the authors initially provided. From the example, the practical difference does not look important.

jpaillard:
The paper [Reid et al., A Study of Error Variance Estimation in Lasso Regression 2016](https://arxiv.org/pdf/1311.5274) looks close to the problem of estimating this variance. The authors recommend using an estimator (eq 3) that does NOT use a correction, saying that:
 > Our simulation results suggest that this estimator is robust to changes in signal sparsity and strength, more so than its competitors

examples/plot_D0CRT_comparison.py
```diff
@@ -0,0 +1,255 @@
+"""
+Impact of the alpha term in d0CRT X-residual variance
+======================================================
+
+As discussed in `#118 <https://github.com/mind-inria/hidimstat/issues/118>`_,
```
bthirion: I don't think that we should keep that in an example. This is only historical.

examples/plot_D0CRT_comparison.py
```diff
@@ -0,0 +1,255 @@
...
+from hidimstat.distilled_conditional_randomization_test import _joblib_distill
+
+# %%
+# Patched distill function
+# -----------------------------------------
+# We override ``_joblib_distill`` to remove the alpha term, matching the
+# formula from the paper: ``sigma2 = ||X_res||^2 / n``.
+
+
+def _joblib_distill_no_alpha(
```
bthirion: I don't understand: this should be part of the main library. This is too low-level for an example.

bthirion (commented):
I don't really understand the motivation for such an example. IIUC, the matter is more to change the library's default way to compute variance in these models.

jpaillard:
The example is not meant to remain; it was just to share empirical results showing that it does not significantly modify the method's behavior. 
I will remove the example and make the change in the code base.

examples/plot_D0CRT_comparison.py
jpaillard: The whole example will be removed, I just experimented with unittest's mock, which allows mocking a class and surgically modifying one of its functions to show how modification of a given function (here the X-residual variance estimation) changes the behavior.

GrituX:
Why was there a correction in the first place ?

GrituX (approved):

bthirion (commented):
Normally in such a PR you want to include a non-regression test.

bthirion (approved):
LGTM, thx

PR #670, opened 2026-06-04
[MAINT] remove poosh dependency upper-bound

jpaillard:
We used to have problems loading the somato dataset from mne-python. This should be solved in the latest release of mne-python: https://github.com/mne-tools/mne-python/pull/13630

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,3 +21,4 @@ under development (0.3.2.dev0)
 - :bdg-primary:`Doc` add naming conventions for classes, files, and functions, as well as citation conventions (:gh:`647` and :gh:`648` by `Marc Hulcelle`_).
 - :bdg-danger:`Fix` Fix typo in issue template (:gh:`659` by `Joseph Paillard`_).
 - :bdg-secondary:`Maint` temporary fix for the CI upper-bounding scikit-learn to 1.9.0 (:gh:`669` by `Joseph Paillard`_).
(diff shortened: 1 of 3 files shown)


GrituX:
IIUC, this means that the current code only supports the latest version of MNE ?

bthirion (approved):
Happy to see that go away.

jpaillard:
mne is only needed for building the documentation, and I think the doc build indeed uses the latest release of the listed dependencies.

GrituX:
Ok, then I guess the `pyproject.toml` file should be updated :
https://github.com/mind-inria/hidimstat/blob/9229c0d701a898b467f6abcd46182220a2eb1dc6/pyproject.toml#L60

Are there other methods that are impacted by the version update ?

jpaillard:
This PR does not change the version of mne we depend on. It just removes the patch we added to load the somato dataset because of a bug introduced in a previous version of MNE. Only the example modified in this PR uses MNE. 

Regarding the .toml I actually think that all the minimum dependencies listed belloe [doc] could be removed since we use https://github.com/mind-inria/hidimstat/blob/9229c0d701a898b467f6abcd46182220a2eb1dc6/tools/documentation/circleci/setup_virtual_environment.sh#L12
which will use the default resolution with the latest versions.

GrituX:
Ok, sorry for the misunderstanding.

GrituX (approved):

PR #671, opened 2026-06-05
[FIX] Speed up library

### LINKED ISSUE #646: Speed up library
We should use Pytorch learners to make the fit of `y` (much) more efficient.

GrituX:
There were several instances of copy operations on X when only slice views were needed. I made the changes using masks instead of a series of `np.copy` and `np.delete` operations.

### DIFF
--- a/src/hidimstat/_utils/utils.py
+++ b/src/hidimstat/_utils/utils.py
@@ -37,6 +37,18 @@ def _check_vim_predict_method(method):
         )
 
 
+def _generate_mask(array_size, indexes, select_indices=True):
+    """
+    Generates a mask to select indexes or its opposite.
+    """
+    if select_indices:
+        mask = np.zeros(array_size, dtype=bool)
+    else:
+        mask = np.ones(array_size, dtype=bool)
+    mask[indexes] = select_indices
+    return mask
+
+
 def get_fitted_attributes(cls):
     """
     Get all attributes from a class that end with a single underscore
--- a/src/hidimstat/_utils/utils.py
+++ b/src/hidimstat/_utils/utils.py
@@ -37,7 +37,7 @@ def _check_vim_predict_method(method):
         )
 
 
-def _generate_mask_group_mask(array_size, indexes, selected=True):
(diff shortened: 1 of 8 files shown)


src/hidimstat/_utils/utils.py
```diff
@@ -37,6 +37,18 @@ def _check_vim_predict_method(method):
         )
 
 
+def _generate_mask(array_size, indexes, select_indices=True):
```
jpaillard:
```suggestion
def _generate_group_mask(array_size, indexes, complement=False):
```

small suggestions: 
 - title could be more explicit
 - IMO having an argument complement to control whether you return a mask of the group or its complement is clearer. But maybe that's just me.

src/hidimstat/_utils/utils.py
```diff
@@ -37,6 +37,18 @@ def _check_vim_predict_method(method):
...
+    """
+    Generates a mask to select indexes or its opposite.
+    """
+    if select_indices:
+        mask = np.zeros(array_size, dtype=bool)
+    else:
+        mask = np.ones(array_size, dtype=bool)
```
jpaillard: I think you could return `~mask` instead of having an if/else

src/hidimstat/knockoffs.py
```diff
@@ -195,7 +192,7 @@ def fit(self, X, y):
             )
         else:
             self.estimator_ = clone(self.estimator)
-        self.estimator_ = seed_estimator(self.estimator_, self.random_state)
+        self.estimator = seed_estimator(self.estimator, self.random_state)
```
jpaillard: To be scikit-learn compatible, we should not modify the attributes set in the constructor. 
Also, that would not seed `self.estimator_` which is the one used.

jpaillard (commented):
This is a good point, thx.

src/hidimstat/_utils/utils.py
GrituX: No you're right. I was struggling to find a naming that made sense, but yours is good. Thx

jpaillard (approved):

PR #679, opened 2026-06-16
[ENH] Adding LOCI

### LINKED ISSUE #678: [ENH] Implement LOCI

- [x] Yes

### Please describe the feature that you would like to see improved or added.

Add an implementation of Leave One Covariate In. 

### Describe the solution you would like to see implemented.

The idea is to have an estimator that computes the response $Y$ for any model from the individual features $X^j$. 
Then compare the loss of that univariate model to a "null model," which would be something like a constant: the mean in regression and class prevalence for classification. 

### Describe alternatives you have considered.

_No response_

_No response_

GrituX:
Initial implementation of the Leave-One-Covariate-In (LOCI) method for feature importance.

Since it's similar to LOCO in spirit, I repurposed the LOCO example and tests to LOCI. However, LOCI doesn't currently pass the performance tests, meaning that it fails to detect most important features in the univariate simulated data.
I've rechecked the code multiple times, and I can't seem to find any code issue. But maybe I'm missing something ?


src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,340 @@
...
+        assert self.pvalues_.shape[0] == y_pred.shape[0], (
+            "The statistical test doesn't provide the correct dimension."
+        )
+        return self.importances_
+
+    def _joblib_fit_one_features_group(
+        self, estimator, X, y, key_features_group
+    ):
+        """
+        Fit the estimator after removing a group of covariates.
```
jpaillard:
```suggestion
        Fit the estimator on a group of covariates.
```

jpaillard (commented):
## Implementation
Regarding the implementation, I would find it more intuitive to define LOCI for a feature $j$ as,

$$
LOCI(j) := \mathcal{L}(f(\emptyset), Y) - \mathcal{L}(f(X_j), Y) 
$$

So that a more predictive feature has higher importance, whereas currently, a more predictive feature leads to lower $\mathcal{L}(f(X_j), Y)$ and thus lower importance. 

We could define $\mathcal{L}(f(\emptyset), Y)$ as the MSE when using the mean of $Y$ as a prediction for regression and the log-loss when predicting the class frequencies for classification. 

Also, ping @AngelReyero WDYT? 

## Tests
For the test, I think we don't want to test for the FDP since LOCI measures marginal importance. It is expected to give >0 importance to features that are simply correlated but not part of the data-generating mechanism (aka the features in `important_features` returned by `data_generator` in the test)

Also, since correlated features get marginal importance from LOCI, if `important_features` corresponds to the data-generating process (conditional importance), it could be that actually `importance[important_features].mean() < importance[non_important_features].mean()` for example, if you have many features highly correlated with one that has a large coefficient in the data-generating mechanism.

AngelReyero:
> Regarding the implementation, I would find it more intuitive to define LOCI for a feature j as,
> 
> L O C I ( j ) := L ( f ( ∅ ) , Y ) − L ( f ( X j ) , Y )

Yes, I do agree with this definition of LOCI (it is the same used in Ewald et al. 2024 and the one we used in Reyero-Lobo et al. 2025).

> So that a more predictive feature has higher importance, whereas currently, a more predictive feature leads to lower L ( f ( X j ) , Y ) and thus lower importance.
> 
> We could define L ( f ( ∅ ) , Y ) as the MSE when using the mean of Y as a prediction for regression and the log-loss when predicting the class frequencies for classification.

I am not sure to understand this second part, but basically this means that we use a model that is using no input, and the optimal one is then the mean for regression (and therefore L ( f ( ∅ ) , Y ) is just the variance) and for classification to always predict with the most voted class (and L ( f ( ∅ ) , Y ) is probability of not belonging to the class). 


- A Guide to Feature Importance Methods for Scientific Inference (Ewald et al. 2024)
- A principled approach for comparing Variable Importance (Reyero et al. 2025)

bthirion:
> For the test, I think we don't want to test for the FDP since LOCI measures marginal importance. It is expected to give >0 importance to features that are simply correlated but not part of the data-generating mechanism (aka the features in `important_features` returned by `data_generator` in the test)

Why ? I think that  it's a good idea in general to test for FDP.

jpaillard:
What I mean is that if we take two variables $X_1, X_2$ with correlation $\rho>0$, even in the case where $Y=X_1$ the LOCI importance of $X_2$ will be greater than 0. So, the FDP test, as it is currently implemented, will not pass. 

I think we could still test for the FDP, if we generate data with $\rho=0$

GrituX:
> Regarding the implementation, I would find it more intuitive to define LOCI for a feature j as,
> 
> L O C I ( j ) := L ( f ( ∅ ) , Y ) − L ( f ( X j ) , Y )
> 
> So that a more predictive feature has higher importance, whereas currently, a more predictive feature leads to lower L ( f ( X j ) , Y ) and thus lower importance.

Ok, that actually makes more sense. I'll change that then.

> For the test, I think we don't want to test for the FDP since LOCI measures marginal importance.
> What I mean is that if we take two variables X 1 , X 2 with correlation ρ > 0 , even in the case where Y = X 1 the LOCI importance of X 2 will be greater than 0. So, the FDP test, as it is currently implemented, will not pass.
> I think we could still test for the FDP, if we generate data with ρ = 0 

I've tried a few setups yesterday with ρ values and indeed the FDP test never passed. I think it's fair and safer to use ρ = 0 for LOCI.

AngelReyero:
> What I mean is that if we take two variables X 1 , X 2 with correlation ρ > 0 , even in the case where Y = X 1 the LOCI importance of X 2 will be greater than 0. So, the FDP test, as it is currently implemented, will not pass.
> 
> I think we could still test for the FDP, if we generate data with ρ = 0

For LOCI the problem is not that we do not control the FDP because of spurious correlation. FDP is just a guarantee that we have and what we are changing is not the guarantee but the null hypothesis, because the LOCI will not give valid p-values for the conditional null hypothesis but for the marginal null hypothesis.

GrituX:
For the multilabel classification case, I computed the y_baseline for the loss reference as the marginal probability of y. Another option is to perform majority voting per label, but I think it makes more sense to have marginal probability. Any thoughts ?

bthirion:
> What I mean is that if we take two variables X 1 , X 2 with correlation ρ > 0 , even in the case where Y = X 1 the LOCI importance of X 2 will be greater than 0. So, the FDP test, as it is currently implemented, will not pass.
> 
> I think we could still test for the FDP, if we generate data with ρ = 0

Or consider that the null variables may indeed be selected by the BH procedure. Nothing wrong with that.

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,352 @@
...
+from sklearn.metrics import mean_squared_error
+
+from hidimstat._utils.docstring import _aggregate_docstring
+from hidimstat._utils.utils import check_statistical_test
+from hidimstat.base_perturbation import BasePerturbation, BasePerturbationCV
+
+
+class LOCI(BasePerturbation):
+    """
+    Leave-One-Covariate-In (LOCO) algorithm
```
jpaillard:
```suggestion
    Leave-One-Covariate-In (LOCI) algorithm
```

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,352 @@
...
+
+    The model is re-fitted for each single feature/group of features. The importance is
```
jpaillard:
```suggestion
    The model is re-fitted on each single feature/group of features. The importance is
```

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,352 @@
...
+        loss: callable = mean_squared_error,
+        statistical_test="ttest",
+        features_groups=None,
+        n_jobs: int = 1,
+    ):
+        super().__init__(
+            estimator=estimator,
+            method=method,
+            loss=loss,
+            n_permutations=1,
```
jpaillard:
```suggestion
```

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,352 @@
...
+        pvalues_ : ndarray of shape (n_groups,)
+            P-values from one-sided t-test testing if importance scores are
+            significantly greater than 0.
+
+        Notes
+        -----
+        The importance score for each group is calculated as the mean increase in loss
+        when that group is perturbed, compared to the reference loss.
+        A higher importance score indicates that perturbing that group leads to
+        worse model performance, suggesting those features are more important.
```
jpaillard: It looks like this is not for LOCI

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,352 @@
...
+    ],
+    """
+    Returns
+    -------
+    selection : ndarray of shape (n_features,)
+        Boolean array indicating selected features (True = selected)
+    importances : ndarray of shape (n_features,)
+        Feature importance scores/test statistics.
+    pvalues : ndarray of shape (n_features,)
+        None because there is no p-value for this method
```
jpaillard: The method returns p-values though 

Also, I think it's OK to say that there are p-values but for the marginal importance

test/test_leave_one_covariate_in.py
jpaillard, line 1: Since LOCI is has a specific "null model" for multiclass classification, can we also add a test for the multiclass classification case?

src/hidimstat/leave_one_covariate_in.py
GrituX: I thought I made all the modifications about the LOCO/LOCI doc mixup, I'm confused. I'll fix that again

examples/plot_loci.py
```diff
@@ -0,0 +1,149 @@
...
+            i - 0.45,
+            i + 0.45,
+            color="tab:olive",
+            alpha=0.3,
+            zorder=-1,
+            label="True Support" if i == 1 else None,
+        )
+ax.legend()
+plt.show()
+
```
bthirion, line 142: There is clearly something fishy with the RF.

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,355 @@
...
+                y_baseline = np.full_like(
+                    y, values[np.argmax(counts)], dtype=int
+                )
+            # For multiclass classification, we take the marginal probability.
+            else:
+                y_baseline = np.full(
+                    (y.shape[0], len(values)), counts / y.shape[0]
+                )
+        elif self.method == "predict":
+            y_baseline = np.full_like(y, np.mean(y), dtype=float)
```
bthirion: This is the culprit: you should instead take the mean on the training set, not the test set.

test/test_leave_one_covariate_in.py
```diff
@@ -0,0 +1,253 @@
...
+from hidimstat._utils.scenario import multivariate_simulation
+from hidimstat.base_perturbation import BasePerturbation
+from hidimstat.statistical_tools.multiple_testing import fdp_power
+
+
+def test_loci():
+    """Test the Leave-One-Covariate-In algorithm on a linear scenario."""
+    X, y, beta, _ = multivariate_simulation(
+        n_samples=150,
+        n_features=200,
```
bthirion: Why do you take so many features ? It uselessly increases computation time

test/test_leave_one_covariate_in.py
```diff
@@ -0,0 +1,253 @@
...
+
+    with pytest.raises(
+        ValueError, match="The estimators require to be fit before to use them"
+    ):
+        fitted_model = LinearRegression().fit(X, y)
+        loci = LOCI(
+            estimator=fitted_model,
+            method="predict",
+        )
+        BasePerturbation.fit(loci, X, y)
```
bthirion, line 174: the error is that loci has not been fitted, right ?

test/test_leave_one_covariate_in.py
```diff
@@ -0,0 +1,253 @@
...
+    assert importance_clf[0].mean() > importance_clf[1].mean()
+
+
+def test_multiclass_loci():
+    """Test LOCI in a multiclass classification setup."""
+    important_features = 4
+    n_features = 20
+
+    X, y = make_classification(
+        n_samples=1000,
```
bthirion: do we need 1000 samples ?

bthirion (commented):
Looks mostly good to me, thx !

test/test_leave_one_covariate_in.py
GrituX: Yes

test/test_leave_one_covariate_in.py
GrituX: We really don't need this much. I quickly checked tests for other methods, and I think we can make the same modifications on several of them.

test/test_leave_one_covariate_in.py
GrituX: No, a hundred is far enough

examples/plot_loci.py
GrituX: Even after the fix of the mean taken on the test set, the problem persists. I'm investigating this.

GrituX: Well, it just seems that RF trained on a single feature has a significantly worse mean MSE than all other models.

bthirion: Yes, but it should not be worse than RF with no feature (when the feature is useful).

GrituX:
The issue with the RF in the example is fixed, now it's ready for a final review.

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        Compute the importance scores for each group of covariates.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            The input samples to compute importance scores for.
+        y : array-like of shape (n_samples,)
+
+        importances_ : ndarray of shape (n_groups,)
```
bthirion, line 125: to be removed  I guess ?
Size is `n_groups` or `n_features` ?

bthirion (approved):
LGTM, thx.

src/hidimstat/leave_one_covariate_in.py
GrituX: Sorry I don't understand, remove what ?
Size is `n_groups`, which equals `n_features` when groups are undefined.

bthirion: this block is repeated below: it is actually the method's return argument.

GrituX: Ok, I fixed it thx !

examples/plot_loci.py
```diff
@@ -0,0 +1,162 @@
+"""
+Leave-One-Covariate-In (LOCI) feature importance with different regression models
+==================================================================================
+
+This example demonstrates how to compare LOCI feature importance [:footcite:t:`Williamson_General_2023`] across different
```
jpaillard, line 5: Is LOCI attributed to Williamson? 
I would rather refer to the scikit-learn implementation of univariate feature selection and explain that LOCI generalizes to non-linear models.

GrituX: No, it was more of a general reference for a paper that explains, among other methods, LOCI.

IIUC, their univariate feature selection is based on an F-test that compares variance, while LOCI studies loss decrease. Both methods study the impact of a single feature, but it seems to me that the methods and underlying metrics are different. I'm not sure I completely understand the link between the two, and whether it makes sense to compare them.

jpaillard: From what I remember, the paper only studies LOCO. 
Maybe we could cite: [A principled approach for comparing Variable Importance](https://arxiv.org/abs/2507.17306) since we follow their description for the implementation.

GrituX: Ok, it makes more sense. I can change the reference. I'll add this change in PR #685.

Issue #297, opened 2025-06-17
Methods to add to Hidimstat


lionelkusch:
This is the list: I will update it following the comment. 
The order will be used as the priority for including new methods.

List of methods to add:
 - [ ] Marginal information/ correlation ([reference](https://link.springer.com/content/pdf/10.1186/1471-2105-9-307.pdf)
 - [ ] PDP: Partial Dependence Plot based Importance ([reference](https://arxiv.org/pdf/1805.04755))
 - [ ] ALE: Accumulate Local Effect based Importance ([reference](https://doi.org/10.1111/rssb.12377))
 - [ ] Friedman’s H-statistic ([reference](https://www.jstor.org/stable/30245114))
 - [ ] DPDP: Differential Partial Dependence Plot  based Importance ([reference](https://arxiv.org/pdf/2210.04542))
 - [ ] DALE: Differential Accumulate Local Effect based Importance ([reference](https://doi.org/10.1111/rssb.12377))Differential
 - [ ] RHALE: Robust and Heterogeneity-aware Accumulated Local Effects based Importance ([reference](https://arxiv.org/pdf/2309.11193))
 - [ ] SAGE: Shapley Additive Global importancE ([reference](https://arxiv.org/pdf/2004.00668))
 - [ ] SHAP : SHapley Additive exPlanations average ([reference](https://link.springer.com/chapter/10.1007/978-3-642-20267-4_3))
 - [ ] DCRT: distilled conditional randomization test [generalize it] ([reference](https://pmc.ncbi.nlm.nih.gov/articles/PMC10323874/))
 - [ ] HRT: Holdout randomization test ([reference](https://arxiv.org/pdf/1811.00645))
 - [ ] CRT: Conditional randomize tests ([referene](https://arxiv.org/pdf/1610.02351))
 - [ ] Knockoff [extensions] ([reference](https://web.stanford.edu/group/candes/knockoffs/papers.html))
 - [ ] Permutation target test ([reference](https://www.nature.com/articles/nrg1521))
 - [ ] Estimator of Permutation target test ([reference](https://pubmed.ncbi.nlm.nih.gov/23285616/))

lionelkusch:
@jpaillard @AngelReyero @bthirion don't hesitate to propose some methods to add and criticise the priority to add this method.

bthirion:
I think we should not try to be exhaistive. Remember that sklearn does not consider a method than has not been cited 200 times.
* I've never cross Firedmann's H statistic ?
[...]

PR #689, opened 2026-06-18
[FEAT] Add ALE (old fork)

abaziree:
Issue #297 

I added 1D ALE for continuous and discrete numeric features, and 2D ALE.
There are currently no tests.

Reference paper: https://arxiv.org/pdf/1612.08468

Use of AI tools:
- [x]  LLM tools were used to generate at least part of the content of this pull request.


src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+
+def _build_quantile_grid(x, grid_resolution):
+    """Build a 1D quantile grid for a single continuous feature column.
+
+    Parameters
+    ----------
+    x : ndarray of shape (n_samples,)
+        Values for one feature.
+    grid_resolution : int
+        Number of bin in the grid.
```
bthirion:
```suggestion
        Number of bins in the grid.
```

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+        return predict_fn
+
+    raise ValueError(
+        "'estimator' must expose at least one of predict_proba, "
+        "decision_function, or predict."
+    )
+
+
+def _build_quantile_grid(x, grid_resolution):
+    """Build a 1D quantile grid for a single continuous feature column.
```
bthirion:
```suggestion
    """Build a 1D quantile grid for a single continuous feature.
```

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+    if grid_resolution <= 0:
+        raise ValueError("'grid_resolution' must be strictly greater than 0.")
+
+    # Use unique values when there are fewer than grid_resolution unique points
+    uniques = np.unique(x[~np.isnan(x)])
+    if uniques.shape[0] <= grid_resolution:
+        return uniques
+
+    probs = np.linspace(0.0, 1.0, grid_resolution + 1)
+    return np.unique(np.percentile(x, probs * 100, method="inverted_cdf"))
```
bthirion: due to the `unique`, the number of values returned can be smaller than `grid_resolution`. Should be indicated in the docstring

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+    Parameters
+    ----------
+    estimator : fitted sklearn-compatible estimator
+        Must expose ``predict``, ``predict_proba``, or ``decision_function``.
+    X : array-like of shape (n_samples, n_features)
+        Training (or evaluation) dataset used to define the quantile grid and
+        to gather local samples in each bin.
+    feature_idx : int
+        Column index of the feature of interest.
+    grid_resolution : int, default=50
```
bthirion: could have an `auto` value, where the number of bins is set in order to minimize the histogram error.

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+def compute_ale_1d_discrete(estimator, X, feature_idx):
+    """Compute the 1D Accumulated Local Effect for a single discrete feature.
+
+    For each value of the feature, the local effect is estimated as the average
+    difference in model output when the feature moves from the lower to the upper
+    neighbour values across all samples that have that value. The cumulative sum
+    of these average effects gives the (uncentered) ALE curve, which is then
+    centered so its weighted mean is zero.
```
bthirion: Is there a reference to justify the behavior ?

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+    bins.  The resulting surface captures the joint effect that cannot be
+    explained by either 1D ALE alone.
+
+    Parameters
+    ----------
+    estimator : fitted sklearn-compatible estimator
+    X : array-like of shape (n_samples, n_features)
+    feature_indices : tuple or list of two ints
+        Column indices ``[i, j]`` of the two features of interest.
+    grid_resolution : int, default=20
```
bthirion: could also have an `auto` mode

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,287 @@
...
+
+    .. math::
+
+        \hat{f}_{j,\text{ALE}}(x) =
+            \int_{z_{0,j}}^{x}
+            \mathbb{E}\!\left[
+                \frac{\partial f(X)}{\partial X_j}
+                \;\middle|\; X_j = z
+            \right] dz - c
+
```
bthirion, line 534: We should include a bilbiographic reference to the definition

bthirion (commented):
I'm still wondering what we should do for discrete features. Isn't PDP a more natural solution for such feature ? I'm worried about the  arbitrary ordering, that may lead to a biased representation.

To be discussed again.

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+            if feature_names is not None:
+                X = pd.DataFrame(X, columns=feature_names)
+            return estimator.predict(X).ravel()
+
+        return predict_fn
+
+    raise ValueError(
+        "'estimator' must expose at least one of predict_proba, "
+        "decision_function, or predict."
+    )
```
GrituX: This function has a few issues :
- L38-54 can be refactored with `getattr(estimator, prediction_func)(X).ravel()`
- imho, nested functions reduce code readability. Here, the function could directly return the prediction for instance

src/hidimstat/visualization/_compute_accumulated_local_effects.py
```diff
@@ -0,0 +1,442 @@
...
+    X_hl[:, idx_j] = quantiles_j[bin_idx_j]
+
+    X_hh = X.copy()
+    X_hh[:, idx_i] = quantiles_i[bin_idx_i + 1]
+    X_hh[:, idx_j] = quantiles_j[bin_idx_j + 1]
+
+    # Second-order local effects
+    local_effects = (predict_fn(X_hh) - predict_fn(X_hl)) - (
+        predict_fn(X_lh) - predict_fn(X_ll)
+    )  # shape (n_samples,)
```
GrituX: bin_idx could be a 1D array which could refactor this code with a loop, and reduce the number of copies of X

src/hidimstat/visualization/_compute_accumulated_local_effects.py
abaziree: yes, i specified it in every docstrings

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: I added the Apley and Zhu reference. Should we also add Molnar's one even though I haven't specifically used it ?

src/hidimstat/visualization/_compute_accumulated_local_effects.py
abaziree: Yep, I agree with your second point, it's not really useful here.

src/hidimstat/visualization/_compute_accumulated_local_effects.py
abaziree: The docstring is maybe not precise enough. In an interval ]x_{k-1}, x_k] of the feature of interest , we calculate the local effect by considering on the one hand samples with value x_{k-1} subsequently shifted to the value x_k, and on the other hand samples with value x_k subsequently shifted to the value x_{k-1}. It's exactly the same formula as the one in the Apley & Zhu paper.

bthirion: Please include a ref to that paper.

abaziree:
Kept as draft to have a history of discussions

PR #690, opened 2026-06-18
[BUG] Sklearn 1.9 compatibility

### LINKED ISSUE #686: [BUG]  skelarn LassoCV API

- [x] Yes

### Describe your issue.

The lastest sklearn LassoCV no longer takes `n_alphas` but `alphas`
Hidimstat code is not up to date.
See e.g. https://github.com/mind-inria/hidimstat/blob/main/src/hidimstat/distilled_conditional_randomization_test.py#L138

Note: it is weird that this has not been captured by CI ?

### Reproducing Code Example

```python
Comes from importing hidimstat in recent environment
```

### Error message

```shell
TypeError: LassoCV.__init__() got an unexpected keyword argument 'n_alphas'. Did you mean 'alphas'?
```

```shell
In [4]: sklearn.__version__
Out[4]: '1.9.0'
```

GrituX:
Fixed deprecation of Sklearn `LassoCV.n_alphas` from 1.9 and switched to `alphas`. This fix requires scikit-learn >= 1.7.
Also fixed deprecation of `LogisticRegressionCV `penalty at initialization (for >=v1.8), and fixed by replacing with `l1_ratios=(0,)` as recommended (this is available from v1.7.2).


bthirion (commented):
Sklearn 1.6 is still pretty recent (less than 2 years old). I think it's a reasonable to support both. this is what is done in nilearn.

GrituX:
Ok, I didn't take the time to check the release date of v1.6, but then it seems reasonable to still support it. I'll make the checks for `alphas` since it is not supported in v1.6, but `l1_ratios` is available and behaves similarly in v1.6

test/conftest.py
```diff
@@ -9,7 +9,7 @@
 
 from hidimstat._utils.scenario import multivariate_simulation
 
-SKLEARN_LT_1_6 = parse(sklearn_version).minor < 6
+SKLEARN_LT_1_6 = parse(sklearn_version).minor <= 6
```
bthirion: Maybe defin it once and for all in a utils file ?

bthirion (commented):
LGTM. Can you make checks on different sklearn versions ?

examples/plot_dcrt_logit.py
```diff
@@ -77,7 +77,7 @@
     # Fit the dCRT-logit model
     dcrt_logit = D0CRT(
         estimator=LogisticRegressionCV(
-            penalty="l1",
+            l1_ratios=(1,),
```
jpaillard, line 81: In sklearn 1.6, this will be ignored and overridden by the default `penalty='l2'` setting. 
https://scikit-learn.org/1.6/modules/generated/sklearn.linear_model.LogisticRegressionCV.html

examples/plot_knockoffs_wisconsin.py
```diff
@@ -102,7 +105,7 @@
 
 lasso_noisy = LogisticRegressionCV(
     Cs=np.logspace(-3, 3, 10),
-    penalty="l1",
+    l1_ratios=(1,),
```
jpaillard, line 110: same compatibility issue as above for sklearn 1.6

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -15,13 +15,32 @@
...
 from hidimstat._utils.utils import (
+    SKLEARN_LT_1_6,
     _check_vim_predict_method,
     check_random_state,
     seed_estimator,
 )
 from hidimstat.base_variable_importance import BaseVariableImportance
 
 
+def _default_lasso_screening():
```
jpaillard: What about having a helper function in utils that handles the argument change with versions? It would take as argument a sklearn estimator and kwargs to allow changing the parameters and handle the switch between versions. 

That would avoid having multiple functions (one for distillation, one for lasso screening, and one that is going to be needed for the Logistic regression ... )

src/hidimstat/distilled_conditional_randomization_test.py
```diff
@@ -153,8 +172,16 @@ def __init__(
...
+        self.lasso_screening = (
+            _default_lasso_screening()
+            if lasso_screening is None
+            else lasso_screening
+        )
+        self.model_distillation_x = (
+            _default_model_distillation_x()
+            if model_distillation_x is None
+            else model_distillation_x
+        )
```
jpaillard: IIRC, we had the discussion with @Remi-Gau that setting class attributes this way isn't the most sklearn-compatible.  
 - in the init it should be `self.model_distillation_x = model_distillation_x ` (for cloning estimators)  
 - in the .`fit()` something like `self.model_distillation_x_ = ...` (with the trailing underscore)

jpaillard: it was for instance in #590

Remi-Gau:
> Sklearn 1.6 is still pretty recent (less than 2 years old). I think it's a reasonable to support both. this is what is done in nilearn.

it is recommended to support sklearn 1.6 till at least the end of the year according to the SPEC0: https://scientific-python.org/specs/spec-0000/

Remi-Gau:
note that there is a sklearn package that can help with some sklearn deprecations: https://github.com/sklearn-compat/sklearn-compat

not checked if it can help in this case, but worth keeping on the radar

examples/plot_knockoffs_wisconsin.py
Remi-Gau: maybe I am missing something but it seems the test with sklearn 1.6 are passing

src/hidimstat/distilled_conditional_randomization_test.py
Remi-Gau: yup sklearn standards is that there should be no logic in your constructor

Remi-Gau: I should probably get back to ensure all estimators are checked for their sklearn compliance: it can avoid headaches down the line

examples/plot_knockoffs_wisconsin.py
GrituX: it's not technically an issue. Instead of taking it into account, sklearn just ignores it, and so instead of applying an l1 correction as requested, sklearn applies a default l2 correction and no warning is issued. I'll replace the penalty with "elasticnet" so that it's taken into account.

jpaillard: Yes exactly, the test would pass but the estimator would silently use L2 instead of L1 penalty

src/hidimstat/distilled_conditional_randomization_test.py
GrituX: indeed, it's much better this way, thanks for the suggestion.

src/hidimstat/_utils/utils.py
```diff
@@ -3,10 +3,37 @@
...
 from scipy.stats import ttest_1samp, wilcoxon
+from sklearn import __version__ as sklearn_version
 
 from hidimstat.statistical_tools.nadeau_bengio_ttest import nadeau_bengio_ttest
 
+SKLEARN_LT_1_6 = parse(sklearn_version).minor <= 6
+SKLEARN_LT_1_9 = parse(sklearn_version).minor == 9
+
+
+def make_sklearn_estimator(estimator_cls, **kwargs):
```
jpaillard: Could you add a test for this function?

GrituX: yes, I'm working on it !

GrituX:
Apparently Codecov doesn't test sklearn at v1.6 (and it's something we can't simulate really), so it delcares that added code is not tested, while the tests are there.

src/hidimstat/_utils/utils.py
bthirion: +a short docstring

test/test_knockoff.py
```diff
@@ -288,6 +289,35 @@ def test_model_x_knockoff_null():
     assert not selected.any()
 
 
+def test_lasso_estimator_alphas():
+    """Test configuration of alphas when estimator is LassoCV depending on SKLEARN version."""
```
bthirion:
```suggestion
    """Test configuration of alphas when estimator is LassoCV depending on Scikit-Learn version."""
```

bthirion (approved):
LGTM, thx.
Simply check with both sklearn versions...

jpaillard:
That's strange because the test-minimal workflow installs sklearn 1.6 
https://github.com/mind-inria/hidimstat/actions/runs/28086017681/job/83152140916#step:5:1

GrituX:
Yes indeed. When I run tests locally with sklearn 1.6, the tests cover what codecov indicates as missing. So I'm a little confused here as to how Codecov computes test coverage.

bthirion:
> Yes indeed. When I run tests locally with sklearn 1.6, the tests cover what codecov indicates as missing. So I'm a little confused here as to how Codecov computes test coverage.

Yeah, I would not worry too much about codecov. It's weird indeed.

PR #692, opened 2026-06-19
[MAINT] fix test for conditional sampler 

### LINKED ISSUE #691: [MAINT] CI sometimes fails with test_conditional_sampling.py::test_group_case_binary_case

- [x] Yes

When pushing changes, tests sometimes fail because of the following line :
https://github.com/mind-inria/hidimstat/blob/60c41941d4feebf769cb55b7d83b3231aec9bf2d/test/samplers/test_conditional_sampling.py#L216
as it results in `assert 0.96 > 0.961` (assertion failures never go above 0.961).
How were these bounds chosen ?

### Suggested fix:

The chosen bounds should be inspected.

_No response_

jpaillard:
- Remove the upper bound on the accuracy of the sampled 
 - Check that the generated samples are not all the same instead


GrituX (approved):

bthirion (approved):
Thx.

PR #699, opened 2026-06-23
[MAINT] allow run CI of workflow_call and pull_request_target

Remi-Gau:
- should fix the failures to run CI on PRs opened from a fork
  https://github.com/mind-inria/hidimstat/pull/599#issuecomment-4766936715

### DIFF
--- a/.github/workflows/call_publish_result.yml
+++ b/.github/workflows/call_publish_result.yml
@@ -100,6 +100,8 @@ jobs:
     steps:
     # initialize the reposititory
     - uses: actions/checkout@v7
+      with:
+        allow-unsafe-pr-checkout: true
 
     # get coverage only for 1 job
     - name: Get data
--- a/.github/workflows/call_test_minimal.yml
+++ b/.github/workflows/call_test_minimal.yml
@@ -38,6 +38,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          allow-unsafe-pr-checkout: true
 
         # skip tests
       - name: Skip the tests
--- a/.github/workflows/call_test_package.yml
+++ b/.github/workflows/call_test_package.yml
@@ -38,6 +38,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          allow-unsafe-pr-checkout: true
 
         # skip tests
       - name: Skip the tests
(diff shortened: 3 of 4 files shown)


Remi-Gau:
pretty sure this needs to be merge to take effect

Remi-Gau:
Setting as draft because there is a potential security risk to do this, so this probably should be discussed first

https://github.com/actions/checkout#whats-new

Remi-Gau:
OK I am going to merge this and check that this allows to run CI on PR from forks: if it works I will then open an issue where to discuss the security risk.

Remi-Gau:
ha this needs at least one approval: up to you @jpaillard @GrituX @bthirion

jpaillard:
From what I understand, this moves back to the previous state of the CI, before checkout v7. 

Do you know if the vulnerability is serious?  From what I understand, the secrets Codecov/CircleCI can be exposed. 

Would there be a way to separate Codecov and CircleCI executions for forks?  It could, for instance, be triggered by a maintainer reviewing the PR.

Remi-Gau:
manual triggers will be a pain in the neck

the nuclear option would be to make sure we do not use workflow_call and pull_request_target but that would require a refactoring of the CI

workflow_call are useful when you need to modularize some steps in CI so you easily perform that same step when running tests or building the doc...

but there are ways around it without having to copy paste entire section of yaml

bthirion (approved):

GrituX:
After a bit of reading, the security risks are the following :
- since the code is executed on the base repo with elevated access (same as maintainer), an attacker's code could abuse write-access and push commits, open/merge PRs, modify repo settings, create releases etc.
- the workflow gets access to the default branch cache, so malicious code can poison it. The cache is for python environments in our case, so I think the risks are somewhat limited ?

From what I understand, what we could risk is losing access to the repository or having it entirely deleted at worst during a malicious attack. The probability of it happening should be pretty low.

The latest recommendations are to switch from `pull_request_target` to `pull_request` as the latter doesn't give elevated access during workflow execution, but it's not clear to me how much needs to be changed.

jpaillard:
The worst case looks pretty bad, right? 

It probably justifies the "nuclear option" of refactoring the CI, WDYT?

bthirion:
> The worst case looks pretty bad, right?
> 
> It probably justifies the "nuclear option" of refactoring the CI, WDYT?

Yes, we should go toward safe options.

GrituX:
I also agree. I will make this my priority as of next week.

Remi-Gau:
superseded by #737

jpaillard:
Should we close it then? @GrituX

GrituX:
Yes, we can still re-open it later if necessary.

PR #701, opened 2026-06-24
[FEAT] Add ALE

### LINKED ISSUE #297: Methods to add to Hidimstat

abaziree:
Issue #297 

I added 1D ALE for continuous and discrete numeric features, and 2D ALE.
There are currently no tests.

Reference paper: https://arxiv.org/pdf/1612.08468

Use of AI tools:
- [x]  LLM tools were used to generate at least part of the content of this pull request.


abaziree:
I added tests for PDP because Codecov was complaining :c
And he doesn't seem to detect the ALE file

bthirion:
Let us know when you need another round of reviews.

abaziree:
> Let us know when you need another round of reviews.

I think it's useful now.

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+# %%
+# Accumulated Local Effects for a Single Discrete Feature
+# -------------------------------------------------------
+# Let's see how ALE handles discrete features. We will create a synthetic
+# dataset where one feature takes on distinct integer values.
+#
+# Unlike continuous features where local differences are evaluated over quantile
+# bins, for discrete features, the local effect is calculated by shifting the
```
bthirion: This is not super clear.

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+n_samples = 1000
+
+discrete_feature = generator.integers(1, 6, size=n_samples)
+continuous_feature = generator.normal(n_samples)
+
+X_discrete = pd.DataFrame(
+    {"Feature_A": discrete_feature, "Feature_B": continuous_feature}
+)
+
+y_discrete = (
```
bthirion:
```suggestion
y = (
```

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,113 @@
...
+# containing a deliberate statistical trap:
+#
+# 1. We generate two features, :math:`X_1` and :math:`X_2`, which are strongly correlated.
+# 2. We define the target variable :math:`y` to depend only on :math:`X_1` via a quadratic function:
+#
+# .. math::
+#
+#   y = X_1^2 + \varepsilon
+#
+# **The Goal:** Because :math`y` has absolutely connection to :math:`X_2`, an ideal
```
bthirion:
```suggestion
# **The Goal:** Because :math:`y` has absolutely connection to :math:`X_2`, an ideal
```

test/visualization/test_partial_dependence_plot.py
```diff
@@ -0,0 +1,66 @@
...
+ale_ids = [p[0] for p in parameters_pdp]
+ale_values = [p[1:] for p in parameters_pdp]
+
+
+@pytest.mark.parametrize(
+    "n_samples, n_features, support_size, rho, seed, value, signal_noise_ratio, rho_serial",
+    ale_values,
+    ids=ale_ids,
+)
+class TestPDP:
```
bthirion: Sorry, can we avoid classes in tests ? I prefer to have them has flat functions, given that the class structure does not bring anything.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+        prediction_function = "decision_function"
+    elif hasattr(estimator, "predict"):
+        prediction_function = "predict"
+    else:
+        raise ValueError(
+            "'estimator' must expose at least one of predict_proba, decision_function, or predict."
+        )
+
+    if feature_names is not None:
+        X = pd.DataFrame(X, columns=feature_names)
```
bthirion, line 53: What if X is already a DataFrame ?

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+
+    pred = getattr(estimator, prediction_function)(X)
+
+    if (hasattr(estimator, "classes_") and len(estimator.classes_) > 2) or (
+        pred.ndim == 2 and pred.shape[1] > 2
+    ):
+        raise ValueError("Multiclass models are not supported.")
+
+    # Binary: keep only the positive-class column
```
bthirion, line 62: Given that, the generalization to multi-class could be handled as a multi-output problem ? (just wondering, and in any case, not in this PR)

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+        Number of bins used to build the quantile grid. Set by default to "auto".
+
+        - If "auto", the number of bins is determined automatically
+          to minimize the histogram error.
+        - Note that the final number of bins in the quantile grid may be
+          strictly less than `grid_resolution` (or the auto-calculated value)
+          if the data contains many duplicate values or fewer unique points
+          than requested.
+
+    confidence_interval : bool, default=False
```
bthirion: I'd be in favor of removing this argument and always compute CIs.
This does not ass too much computation, right ?

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+    feature_idx,
+    confidence_interval=False,
+    confidence_level=0.95,
+    percentiles=(0.05, 0.95),
+):
+    """Compute the 1D Accumulated Local Effect for a single discrete feature.
+
+    For each value of the feature, the local effect is estimated as the average
+    difference in model output when the feature moves from the lower to the upper
+    neighbour values across all samples that have that value. The cumulative sum
```
bthirion: this does not sound very clear to me...

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+    Parameters
+    ----------
+    estimator : fitted sklearn-compatible estimator
+        Must expose `predict`, `predict_proba`, or `decision_function`.
+    X : array-like of shape (n_samples, n_features)
+        Training (or evaluation) dataset used to define the quantile grid and
+        to gather local samples in each bin.
+    feature_idx : int
+        Column index of the feature of interest.
+    confidence_interval : bool, default=False
```
bthirion: Same comment here

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+
+
+def compute_ale_1d_discrete(
+    estimator,
+    X,
+    feature_idx,
+    confidence_interval=False,
+    confidence_level=0.95,
+    percentiles=(0.05, 0.95),
+):
```
bthirion: **Sorry if I misunderstand** but you seem to do relatively complex stuff, while a simple call to boxplot would probably be sufficient ?

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+
+    idx_i, idx_j = feature_indices
+    x_i, x_j = X[:, idx_i], X[:, idx_j]
+
+    quantiles_i = _build_quantile_grid(
+        x_i, grid_resolution=grid_resolution, percentiles=percentiles
+    )
+    quantiles_j = _build_quantile_grid(
+        x_j, grid_resolution=grid_resolution, percentiles=percentiles
+    )
```
bthirion, line 328: as  a side not, the grid should normally be a bit coarser when it's 2D. But can be addressed in a forthcoming PR.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+    References
+    ----------
+    .. footbibliography::
+    """
+
+    def __init__(self, estimator, feature_names=None):
+        self.estimator = estimator
+        self.feature_names = feature_names
+
+    def _resolve_feature_type(self, X, feature, feature_type):
```
bthirion: If I'm not mistaken there is already a function in the library that does that. please use it to avoir reimplementing

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1062 @@
...
+        """Compute and display the ALE plot for one or two features.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            Dataset used to build the quantile grid and to gather local effects.
+        features : int or list of int
+            Feature index (1D ALE) or pair of feature indices (2D ALE).
+        feature_type : string among "auto", "discrete", "continuous", or "categorical"
+            Specify the type of values the feature has for 1D ALE. Set by default to auto and in this case :
```
bthirion: You should not write line longer than 80 chars

bthirion (commented):
Thx this looks good, in particular the plots are nice.
I think I did not understand the discrete case well.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Yes by doing an ALE plot for each class.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: It's nearly nothing in computation. I think it remains useful to have the option to display confidence intervals or not. Perhaps we could set the default value to "True" instead.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: I don't really understand why mentionning boxplot. Here we just compute the ALE like in the continuous case but instead of using quantiles to create the bins we use the unique values of the feature since it is discrete (for instance a feature years). Therefore the bins used are ]x_k, x_{k+1}] with x_0, x_1, ..., x_n the unique values of the feature of interest i.
Since the local effect of the k-th bin correspond to the average calculated over each points z in bin  [x_k, x_{k+1}] of the differences f(x_{k+1}, z_{-i}) - f(x_k, z_{-i}) according to the equation (15) of https://arxiv.org/pdf/1612.08468, and because the value of the i-th feature is either x_k or x_{k+1} in this bin, we compute the differences when the value shift up/down (resp X_high and X_low) and average them.

I hope that is clearer.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Explained in the answer about boxplots. I have this proposition : 
 """Let x_0, x_1, ..., x_n the unique values of the feature of interest. For each bin [x_k, x_{k+1}], the local effect is estimated as the average difference in model output when the feature moves from the lower to the upper bin edge (i.e. x_k and x_{k+1}) across all samples that fall within that bin. The cumulative sum of these average effects gives the (uncentered) ALE curve, which is then centered so its weighted mean is zero."""

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: I don't really understand the issue.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: yes and i think it could be interesting to add a condition such as len_unique_values / len_values <= 0.001 in the current function (_check_data_type in samplers/conditional_sampling .py) to consider a feature as categorical

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: In our usage, X is always a np.ndarray but in general if X is a DataFrame it just returns X !

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+# The 2D ALE isolates pure second-order interaction effects. Here, we plot the
+# interaction between `HouseAge` and `MedInc`.
+#
+# How to Read the Plot:
+# ^^^^^^^^^^^^^^^^^^^^^
+# Because main effects are completely removed, the values on this plot show only
+# the effect from the correlation between the two variables.
+#
+# - A value of 0 means the two features act completely independently (their
+#   combined effect is just the sum of their individual 1D parts).
```
bthirion:
```suggestion
#   combined effect is just the sum of their individual parts).
```

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+# - Positive or negative areas indicate that the combined effect is stronger or
+#   weaker than what the individual 1D curves would imply.
```
bthirion:
```suggestion
#   weaker than what the individual curves would imply.
```

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+)
+plt.show()
+
+
+# %%
+# Accumulated Local Effects for a Single Discrete Feature
+# -------------------------------------------------------
+# Let's see how ALE handles discrete features. We will create a synthetic
+# dataset where one feature takes on distinct integer values.
+#
```
bthirion, line 114: Actually, the point is not that features are  integer, but that they be discrete.

bthirion: Rather than simulating data isn't it possible to use a discrete variable in a public dataset or discretize a continuous variable. It would remove some boilerplate.

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,116 @@
...
+================================
+
+In this notebook, we compare **Partial Dependence Plots (PDP)** and
+**Accumulated Local Effects (ALE)**.
+
+The Problem with PDP
+^^^^^^^^^^^^^^^^^^^^
+PDPs work by marginalizing over the distribution of the features. If
+features :math:`X_1` and :math:`X_2` are highly correlated, the PDP will calculate
+predictions for points that are physically impossible (e.g. a 100 :math:`\\mathrm{m}^2`
```
bthirion:
```suggestion
predictions for points that are impossible (e.g. a 100 :math:`\\mathrm{m}^2`
```

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,116 @@
...
+plt.show()
+
+
+# %%
+# Conclusion
+# ^^^^^^^^^^
+# - The PDP shows a prominent U-shaped parabola for :math:`X_2`. An analyst looking at
+#   this plot would falsely conclude that increasing or decreasing :math:`X_2` directly
+#   increases the target variable. This is a severe error.
+# - ALE correctly shows a mostly flat line (near zero), indicating that :math:`X_2` has
```
bthirion: If it is possible to have error estimates, the result would be clearer. Could be in a forthcoming PR though.

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: OK. 
Note that a DataFrame is really the kind of dtype people we use in practical situations. Maybe this should be better reflected in docstrings.

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: But here we raise an error ? Sorry if I misunderstood

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+        Number of bins in the grid. Set by default to "auto".
+
+        - If "auto", the number of bins is determined automatically
+          to minimize the histogram error.
+        - Note that the final number of bins in the returned grid may be
+          strictly less than `grid_resolution` (or the auto-calculated value)
+          if the data contains many duplicate values or fewer unique points
+          than requested.
+
+    percentiles : tuple of float, default=(0.05, 0.95)
```
bthirion: Why not give directely percentiles in a scale from 0 to 100 ? This is what the name actually means...

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+
+
+def compute_ale_1d_continuous(
+    estimator,
+    X,
+    feature_idx,
+    grid_resolution="auto",
+    confidence_interval=True,
+    confidence_level=0.95,
+    percentiles=(0.05, 0.95),
```
bthirion: See my comment above on percentiles

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+def _bin_indices(x, quantiles):
+    """Assign each sample to a bin defined by *quantiles*.
+
+    Samples are placed in bin `k` when `quantiles[k] <= x < quantiles[k+1]`.
+    The last bin is closed on the right. Indices are clipped so that every
+    sample falls within `[0, len(quantiles) - 2]`.
+
+    Parameters
+    ----------
+    x : ndarray of shape (n_samples,)
```
bthirion: For a "private" function it's OK not to put full docstrings, but if you create one with parameters etc., I think it's better to indicate the role of each input.

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: OK, sorry, I misunderstood.

bthirion: Now, my question is, the distinction between continuous and discrete is almost a detail. Maybe refactor the code to handle both together ?

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: Yeah, I like that one.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+        Whether to compute the confidence intervals of the ALE curve.
+    confidence_level : float, default=0.95
+        The confidence level used to compute the confidence intervals (e.g., 0.95 for 95%).
+    percentiles : tuple of float, default=(0.05, 0.95)
+        The lower and upper percentile used to create the extreme values for the grid.
+        Must be in [0, 1].
+
+    Returns
+    -------
+    result : dict with the following keys:
```
bthirion: Why not directly have ale, quantiles and ale_err as outputs. Nesting them in a dict makes the result less transparent.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
```
bthirion: Why not directly have ale, quantiles and ale_err as outputs. Nesting them in a dict makes the result less transparent.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+          than requested.
+
+    percentiles : tuple of float, default=(0.05, 0.95)
+        The lower and upper percentile used to create the extreme values for the grid.
+        Must be in [0, 1].
+
+
+    Returns
+    -------
+    result : dict with the following keys:
```
bthirion: some comment on result

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: If you have B bins per axis, you'll have B**2 bins in a 2D grid. This means that you'll have less data to compute effects and thus more noisy plots.
A fix is to use B' < B bins per axis so that B' ** 2 is not too large.
The formula for B' exists. LMK if you don't find it.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+
+    def plot(
+        self,
+        X,
+        features,
+        feature_type="auto",
+        grid_resolution="auto",
+        confidence_interval=True,
+        confidence_level=0.95,
+        percentiles=(0.05, 0.95),
```
bthirion: see my comment on percentiles

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+        ax_right.set_ylabel(
+            "Mean modele prediction", rotation=270, labelpad=15
+        )
+
+        sns.despine(ax=ax_main, right=False)
+        plt.tight_layout()
+        return axes
+
+    @staticmethod
+    def _plot_1d_discrete(
```
bthirion: isnt'it possible to refactor a bit: continuous and discrete cases only have very minor difference.

bthirion (commented):
Thx a lot ! it's much better already.
What if a user requests ALE on non-numerical features ?

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Ok I will mention it even if this function is not intended to be used by people

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Currently it's the case

examples/plot_accumulated_local_effects.py
abaziree: > Actually, the point is not that features are integer, but that they be discrete.

Yes I mixed everything, it's not very accurate here

> Rather than simulating data isn't it possible to use a discrete variable in a public dataset or discretize a continuous variable. It would remove some boilerplate.

I will try to find that

examples/plot_comparison_pdp_ale.py
abaziree: Sorry if I misunderstand but do you mean the certitude intervals ? For ALE it's indeed possible but for PDP there is no such option. I can try to add bootstrap for PDP in the PR #703 if it concerns that.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: I took the same convention as for the `percentiles` parameter in scikit-learn's PDPs, since that is the one used by the library. I will change that

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Inspired by the PDP implementation which also returns a dict and I find it's easier to use. Should we change that here ?

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Same answer

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Same answer

examples/plot_accumulated_local_effects.py
GrituX: Isn't AveRooms from the Cali Housing dataset a discrete one ? Maybe we can plot the ALE on this one.

abaziree: No because it has been made per block group (i.e. according to what I've found it is the smallest geographical unit for which the U.S. Census Bureau publishes sample data). Instead to discretize it I choose to take the Tips dataset where the `size` feature is in {1, 2, 3, 4, 5}

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,116 @@
...
+PDPs work by marginalizing over the distribution of the features. If
+features :math:`X_1` and :math:`X_2` are highly correlated, the PDP will calculate
+predictions for points that are physically impossible (e.g. a 100 :math:`\\mathrm{m}^2`
+apartment with 10 bedrooms). This leads to extrapolation bias.
+
+The ALE Solution
+^^^^^^^^^^^^^^^^
+ALE plots, proposed by :footcite:t:`apley2020accumulatedlocaleffects`, calculate the
+model's behavior based on conditional distributions. They only look at how the
+prediction changes when a feature is varied locally, given the values of other features.
```
GrituX:
```suggestion
prediction changes when a feature varies locally, given the values of other features.
```

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,116 @@
...
+pdp = PDP(model, feature_names=X.columns)
+_ = pdp.plot(X_test, features=1)
+
+# 2. ALE Plot
+ale = ALE(model, feature_names=X.columns)
+_ = ale.plot(
+    X_test, features=1, grid_resolution=100, confidence_interval=False
+)
+
+plt.show()
```
GrituX, line 105: It would be nice to make the PDP y axis coincide with the ALE "mean model prediction" y axis, if possible.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1053 @@
...
+
+    if feature_names is not None:
+        X = pd.DataFrame(X, columns=feature_names)
+
+    pred = getattr(estimator, prediction_function)(X)
+
+    if (hasattr(estimator, "classes_") and len(estimator.classes_) > 2) or (
+        pred.ndim == 2 and pred.shape[1] > 2
+    ):
+        raise ValueError("Multiclass models are not supported.")
```
GrituX, line 60: In scikit-learn, multiclass is not necessarily 2d. The target classes might very well be 1d something like `[0, 3, 2, 1, 0]` where each number is a target class. 2d arrays for targets generally indicate multilabel / multioutput situations.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: It's clearly possible to refactor the hand with average, cumulative sum and CIs but not the computation of local effects since in the discrete case, values ​​are used in two bins because they always lie on the boundary and it results in the computation of `local_effect_low` and `local_effects_high`, unlike in the continuous case where we only compute `local_effects`. Therefore I did something like that to refactor.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Yep, there is max 5 lines which change between the 2 versions

examples/plot_comparison_pdp_ale.py
abaziree: That seems impossible to me, since the y-axis depends on the respective values ​​of the PDP and ALE which are not necessarily in the same interval. Perhaps it would be possible to pass a Matplotlib axis as a parameter to plot the curve on, in cases where a comparative display is desired.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: You're right. Here we have :
- `hasattr(estimator, "classes_") and len(estimator.classes_) > 2` acts as a first check to catch standard multiclass classifiers (even if their predict method returns a 1D array of labels).
- `pred.ndim == 2 and pred.shape[1] > 2` serves as a fallback to catch multioutput/multilabel scenarios or multiclass probability matrices if the estimator doesn't natively expose the .classes_ attribute.

For multiclass model it's because if a model outputs 1D class labels (e.g., [0, 3, 2, 1, 0]), computing local differences f(x_{k+1}) - f(x_k) would mean subtracting nominal labels (e.g., Class 3 - Class 0 = 3). Mathematically, this would treat categorical targets as continuous variables, which produces meaningless results.

For multi probability output model we could handle that by computing ALE curves for each class in the future.

examples/plot_comparison_pdp_ale.py
GrituX: That should be possible since `**kwargs`is passed to the plotting function.

src/hidimstat/visualization/accumulated_local_effects.py
GrituX: I agree, but I think it's a bit weird to mix different objects or variables when you test something as you may fall into weird edgecases. Another possibility of testing for multiclass classifiers without relying on the "classes_" attribute would be to do something like `pred.ndim == 1 and len(set(pred)) > 2`. It's more costly, but it avoids mixing variables to test something. But since all scikit-learn classifiers should have the "classes_" member, I think it's fine to keep it as is for the moment.

examples/plot_comparison_pdp_ale.py
abaziree: Finally it's doable just by manipulating the returned axes.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: I find something. Because we want N/B ~ N/(B')^2 to have the same density as in 1D in each bin, then we take B' = floor(sqrt(B)).

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,146 @@
...
+# %%
+# Accumulated Local Effects on a Pair of Features
+# -----------------------------------------------
+# The 2D ALE isolates pure second-order interaction effects. Here, we plot the
+# interaction between `HouseAge` and `MedInc`.
+#
+# How to Read the Plot:
+# ^^^^^^^^^^^^^^^^^^^^^
+# Because main effects are completely removed, the values on this plot show only
+# the effect from the correlation between the two variables.
```
bthirion:
```suggestion
# the effect from the interaction between the two variables.
```

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,146 @@
...
+
+_ = ale.plot(
+    X_test,
+    features=[1, 0],
+    grid_resolution=25,
+    percentiles=(0, 100),
+    cmap="viridis",
+)
+plt.show()
+
```
bthirion, line 108: Maybe explain what the level sets mean.

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,121 @@
...
+ale_axes = ale.plot(
+    X_test, features=1, grid_resolution=100, confidence_interval=False
+)
+
+mean_pred = model.predict(X_test).mean()
+pdp_ymin, pdp_ymax = pdp_axes[1].get_ylim()
+ale_axes[1].set_ylim(pdp_ymin - mean_pred, pdp_ymax - mean_pred)
+ale_axes[1].figure.axes[2].set_ylim(pdp_ymin, pdp_ymax)
+
+plt.show()
```
bthirion, line 105: I find the plots a bit small. Can you increase the size ?

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: Do we agree that ALE is only defined for regression (not classification) problems ?

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: Yes, please.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,993 @@
...
+    -------
+    result : dict with the following keys:
+
+        `"ale"` : ndarray of shape (n_quantiles_i, n_quantiles_j)
+            2D ALE values evaluated at each 2D bin corner.
+        `"quantiles_i"` : ndarray of shape (n_quantiles_i,)
+            Bin edges for the first feature.
+        `"quantiles_j"` : ndarray of shape (n_quantiles_j,)
+            Bin edges for the second feature.
+    """
```
bthirion, line 315: check that percentiles[0] < percentiles[1]

bthirion (commented):
We're getting close ! thx,

src/hidimstat/visualization/accumulated_local_effects.py
GrituX: I think that it can work with classification by working with class probabilities. The `iml` (Interpretable ML) package in R-Studio allows ALE to be computed for classification tasks.
But I agree that the formulation is a bit awkward for classification.

abaziree: Yep it's that. For classification problems it's defined when the predictor return a probability and not just a class.
Here I just allow it for binary predictor but for a multiclass predictor it could be possible to return the ALE curves of each class (based on the predicted proba)

examples/plot_comparison_pdp_ale.py
abaziree: Only in this example or in general ? I tried to do it but when compiling the doc, plots are always compressed to fit side by side. I don't know if there is another solution than creating two different code cells which plot both curves to solve that. Maybe there is the solution of adding an axe parameter in the plotting function but I think it should be done in another PR if wanted and that is a bit overkill for that.

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+# Because main effects are completely removed, the values on this plot show only
+# the effect from the interaction between the two variables.
+#
+# - A value of 0 means the two features act completely independently (their
+#   combined effect is just the sum of their individual parts).
+# - Positive or negative areas indicate that the combined effect is stronger or
+#   weaker than what the individual curves would imply.
+# - The level sets display the iso-proportions of the 2D density of the data,
+#   meaning for instance that 30% of the probability mass lie outside of the contour
+#   drawn for 0.3.
```
bthirion, line 98: This is wrong, because is in that case, the number should increase with the enclosed area.

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: In my current understanding, we should rather limit ALE to legit cases, and develop a better alternative that can deal with classification problems.

src/hidimstat/visualization/accumulated_local_effects.py
bthirion: No, that's not the right heuristic. Will send it to you.

bthirion: https://stats.stackexchange.com/questions/114490/optimal-bin-width-for-two-dimensional-histogram

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1001 @@
...
+
+        ax_cbar_right = ax_cbar.twinx()
+        ymin, ymax = ax_cbar.get_ylim()
+        ax_cbar_right.set_ylim(ymin + mean_prediction, ymax + mean_prediction)
+        ax_cbar_right.set_ylabel(
+            "Mean model prediction", rotation=270, labelpad=15
+        )
+
+        fig.subplots_adjust(left=0.1, right=0.9, bottom=0.16, top=0.9)
+        return axes
```
bthirion, line 871: just wondering. Shouldn't we factor the plotting with PDP by using a base plotting class ?

bthirion (commented):
We're close. Just left a few comments

src/hidimstat/visualization/accumulated_local_effects.py
GrituX: I agree. But I think this should be dealt with in a separate PR.

docs/tools/references.bib
jpaillard, line 1: If possible, put it in alphabetical order. It makes it easier to check for duplicate references.

examples/plot_accumulated_local_effects.py
```diff
@@ -0,0 +1,149 @@
...
+Visualization with Accumulated Local Effects (ALE)
+==================================================
+
+This example demonstrates how to use **Accumulated Local Effects (ALE)**,
+as defined by :footcite:t:`apley2020accumulatedlocaleffects`, to interpret
+machine learning models within the `hidimstat` library.
+
+ALE plots allow you to examine a model's dependence on a single feature or a pair of
+features. Unlike Partial Dependence Plots (PDPs), ALE avoids extrapolation bias when
+features are correlated by averaging localized differences in predictions within conditional intervals.
```
jpaillard:
```suggestion
features. Unlike Partial Dependence Plots (PDPs), ALE avoids extrapolation bias, which occurs when the model is used to predict from out-of-distribution samples,
 by averaging localized differences in predictions within conditional intervals.
```

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1001 @@
...
+import numpy as np
+import pandas as pd
+import seaborn as sns
+from scipy import stats
+from sklearn.utils.validation import check_is_fitted
+
+from hidimstat.samplers.conditional_sampling import _check_data_type
+
+
+def _predict_fn(estimator, X):
```
jpaillard: To be consistent with the rest of the API, I would suggest exposing the prediction method. 
See, for instance, https://github.com/mind-inria/hidimstat/blob/6f3a003035acbfaa66c4e8315fe65f76340e2b0d/src/hidimstat/leave_one_covariate_in.py#L50

If the user wants to plot the decision function, they should be able to do it, even if their estimator has a `predict_proba` method.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1001 @@
...
+        squared_deviations = (combined_effects - sample_means) ** 2
+        sum_squared_deviations = np.bincount(
+            combined_idx, weights=squared_deviations, minlength=n_bins
+        )
+
+        var_of_mean = np.zeros(n_bins, dtype=float)
+        valid_bins = bin_counts > 1
+        var_of_mean[valid_bins] = sum_squared_deviations[valid_bins] / (
+            bin_counts[valid_bins] * (bin_counts[valid_bins] - 1)
+        )
```
jpaillard, line 254: Would it be possible to have something a bit more basic first? Like simply computing  the std per bin: `np.std(combined_effects[bin_mask]) for bin_mask in ...` 

For the CI you suggest, it assumes that the effects follow a normal distribution, which is not so clear. We might want to keep it but maybe not as the default option.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1001 @@
...
+    return ale, grid_values, ale_err
+
+
+def compute_ale_2d(
+    estimator,
+    X,
+    feature_indices,
+    grid_resolution="auto",
+    percentiles=(5, 95),
+):
```
jpaillard, line 271: Would it be possible to move the grid computation out? 

That would make the code more readable and would also allow us to reuse the logic to build the grid in other functions.

examples/plot_comparison_pdp_ale.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+# %%
+# Conclusion
+# ^^^^^^^^^^
+# - The PDP shows a prominent U-shaped parabola for :math:`X_2`. An analyst looking at
+#   this plot would falsely conclude that increasing or decreasing :math:`X_2` directly
+#   increases the target variable. This is a severe error.
```
jpaillard: I would add more nuance here. I think there are two different points in the comparison: 
 1. PDP can lead to extrapolation, like predicting the price of a house in the ocean. This is a problem. Here, ALE is clearly a solution. 
 2. Giving importance to correlated features is **not always** a bad thing. The user just needs to be aware of it. For instance, acting directly on X1 might be impossible, and the only intervention option might be to act on X2 and indirectly modify Y via X1. In this case, you might still want to give X2 importance (like Shapley would). 
So I would modify the phrasing to highlight that ALE is a relevant tool to isolate the unique contribution of a given feature after conditioning on the others. Whether this is desirable depends on the use case.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -0,0 +1,1001 @@
...
+        / 4.0
+    )  # average of the four corner values
+
+    # Centre: subtract the sample-weighted mean (only over occupied bins)
+    ale -= np.sum(ale_centers * bin_counts) / bin_counts.sum()
+
+    return ale, quantiles_i, quantiles_j
+
+
+class ALE:
```
GrituX, line 433: I don't understand why all functions above are outside of the class. To me, they are very related to the behavior of ALE, so they should be included in the class. Why were they excluded ?

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+    class DummyNoPredict(BaseDummy):
+        pass
+
+    class DummyWithFeatureNames(BaseDummy):
+        def __init__(self):
+            super().__init__()
+            self.feature_names_in_ = ["A", "B"]
+
+        def predict(self, X):
+            return np.array([0] * len(X))
```
GrituX: I don't think that nested classes are a good coding practice, even for tests. Can you please move them outside of the function ?

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+        feature_idx=important_features[0],
+        feature_type="continuous",
+        grid_resolution=grid_resolution,
+        confidence_interval=False,
+    )
+    assert isinstance(ale, np.ndarray)
+    assert isinstance(grid_values, np.ndarray)
+    assert ale_err is None
+    assert len(ale) == len(grid_values)
+    assert len(grid_values) <= grid_resolution + 1
```
GrituX, line 189: You do these tests quite often in other functions. I would suggest doing this once for 1d and once for 2d and then being more specific on other tests.

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+        ValueError, match="has fewer than 2 unique quantile edges"
+    ):
+        compute_ale_1d(
+            model,
+            X_const,
+            feature_idx=0,
+            feature_type="continuous",
+            grid_resolution=10,
+        )
+    with pytest.raises(ValueError, match="must be a string among"):
```
GrituX: what does this test ? What does the match string target ?

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+        ValueError, match="'percentiles' must be a tuple of 2 floats"
+    ):
+        compute_ale_1d(
+            model,
+            X,
+            feature_idx=0,
+            feature_type="categorical",
+            percentiles=(-100, 50),
+        )
+    with pytest.raises(
```
GrituX, line 315: Same as above

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+    ):
+        compute_ale_2d(model, X, feature_indices=[0, 1], percentiles=[5, 95])
+    with pytest.raises(
+        ValueError, match="'percentiles' must be a tuple of 2 floats"
+    ):
+        compute_ale_2d(
+            model, X, feature_indices=[0, 1], percentiles=(5, 25, 50, 75, 95)
+        )
+    with pytest.raises(
+        ValueError, match="'percentiles' must be a tuple of 2 floats"
```
GrituX, line 414: Same as above

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+    ):
+        compute_ale_2d(
+            model, X, feature_indices=[0, 1], percentiles=(-100, 50)
+        )
+    with pytest.raises(
+        ValueError, match="'percentiles' must be a tuple of 2 floats"
```
GrituX, line 420: Same as above

test/visualization/test_partial_dependence_plot.py
```diff
@@ -0,0 +1,75 @@
...
+from hidimstat._utils.scenario import multivariate_simulation
+from hidimstat.visualization import PDP
+
+matplotlib.use("Agg")
+
+
+parameters_pdp = [("pdp", 200, 20, 3, 0.3, 92, 1.0, 10.0, 0.0)]
+
+pdp_ids = [p[0] for p in parameters_pdp]
+pdp_values = [p[1:] for p in parameters_pdp]
```
GrituX: It seems that these are not used in the following code. If you want to configure the `mutlivariate_simulation` method with the values that you wrote here, you should have a look at how it's done in `test_conditional_feature_importance.py` for instance.

test/visualization/test_accumulated_local_effects.py
```diff
@@ -0,0 +1,563 @@
...
+    ):
+        compute_ale_1d(
+            model,
+            X,
+            feature_idx=0,
+            feature_type="categorical",
+            percentiles=(-100, 50),
```
GrituX, line 313: Technically -100 is a float. The function should throw a ValueError when one of the percentiles is out of the expected ]0, 100[ range.

GrituX (commented):
Thanks for the implementation ! Here are a few suggestions to improve the code quality

examples/plot_accumulated_local_effects.py
abaziree: Changed with inside instead of outside

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: It could be possible but it's not very useful to modify it now since it is done in #703 to compute CIs with bootstrap.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Yes I think I will move all the logic used to build grids in _utils because it could be used for other things than ALE.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: For me the class is mainly here to plot easily the ALE. I think it is preferable to have these functions outside to allow the user to use them without having to instantiate the class if he wants to manipulate directly the output.

test/visualization/test_accumulated_local_effects.py
abaziree: yup

src/hidimstat/visualization/accumulated_local_effects.py
GrituX: The instatiation cost is not that big, so I don't think it's that big of an issue. But you're right, there's a concept gap between the goal of the class and the goal of this function.

test/visualization/test_accumulated_local_effects.py
abaziree: It checks if `match` is included in the raised error. However the match parameter doesn't accept strings containing characters such as `[`, `]`, `.` . That's why I just put a part of the raised error as `match`.

test/visualization/test_accumulated_local_effects.py
abaziree: Yep I know the full raised error is "'percentiles' must be a tuple of 2 floats in [0, 100] in increasing order". Same reason as above

test/visualization/test_accumulated_local_effects.py
abaziree: Same as above

test/visualization/test_accumulated_local_effects.py
abaziree: Same as above

test/visualization/test_accumulated_local_effects.py
abaziree: Same as above

test/visualization/test_partial_dependence_plot.py
abaziree: Oh I forgot that ! thx

test/visualization/test_accumulated_local_effects.py
GrituX: Yes, but I was more wondering about the string that you are trying to match with. It's hard to identify from your code. Could you paste it here please?

abaziree: It's "must be a string among 'continuous' and 'categorical'". There is no conflict with characters so I put the full error in this match.
It catches this error : https://github.com/mind-inria/hidimstat/blob/b72c1b22957bf9d22a9893cea973183cf8d80451/src/hidimstat/visualization/accumulated_local_effects.py#L217-L219

bthirion (approved):
Great, thx ! Just waiting for @GrituX to agree with all points in the PR.

GrituX (approved):

GrituX:
The code is good to me !

PR #703, opened 2026-06-29
[ENH] Bootstrap for ALE confidence intervals

abaziree:
Related to PR #701 

Instead of calculating the standard deviation for each bin edge based on a single ALE curve, we calculate multiple curves to obtain more reliable confidence intervals.

### DIFF
--- a/src/hidimstat/visualization/accumulated_local_effects.py
+++ b/src/hidimstat/visualization/accumulated_local_effects.py
@@ -4,6 +4,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import seaborn as sns
+from joblib import Parallel, delayed
 from scipy import stats
 from sklearn.utils.validation import check_is_fitted
 
@@ -73,9 +74,11 @@ def compute_ale_1d(
     feature_type,
     method="predict",
     grid_resolution="auto",
+    percentiles=(5, 95),
     confidence_interval=True,
     confidence_level=0.95,
-    percentiles=(5, 95),
+    n_bootstraps=20,
+    n_jobs=-1
 ):
     """Compute the 1D Accumulated Local Effect for a single numerical feature.
 
@@ -110,13 +113,17 @@ def compute_ale_1d(
           if the data contains many duplicate values or fewer unique points
           than requested.
 
+    percentiles : tuple of float, default=(5, 95)
+        The lower and upper percentile used to create the extreme values for the grid.
+        Must be in [0, 100].
     confidence_interval : bool, default=True
         Whether to compute the confidence intervals of the ALE curve.
     confidence_level : float, default=0.95
         The confidence level used to compute the confidence intervals (e.g., 0.95 for 95%).
-    percentiles : tuple of float, default=(5, 95)
-        The lower and upper percentile used to create the extreme values for the grid.
-        Must be in [0, 100].
+    n_bootstraps : int, default=20
+        Number of bootstrap samples to generate if `confidence_interval` is True.
+    n_jobs : int, default=-1
+        Number of jobs to run in parallel during bootstrapping. `-1` means using all processors.
 
     Returns
     -------
@@ -144,23 +151,10 @@ def compute_ale_1d(
                 f"Feature {feature_idx} has fewer than 2 unique quantile edges. Increase grid_resolution or check your data."
             )
 
-        bin_idx = _bin_indices(x, grid_values)
-
-        # For each sample, evaluate the model at the lower and upper edge of its bin
-        X_low = X.copy()
-        X_high = X.copy()
-        X_low[:, feature_idx] = grid_values[bin_idx]
-        X_high[:, feature_idx] = grid_values[bin_idx + 1]
-
-        local_effects = _predict_fn(estimator, X_high, method) - _predict_fn(
-            estimator, X_low, method
-        )  # shape (n_samples,)
-
-        combined_idx = bin_idx
-        combined_effects = local_effects
-        min_weight_bin = 0
+        X_bootstrap = X
 
     elif feature_type == "categorical":
+        # Grid defined by unique values
         if (
             not isinstance(percentiles, tuple)
             or len(percentiles) != 2
@@ -171,7 +165,6 @@ def compute_ale_1d(
                 "in [0, 100] in increasing order"
             )
 
-        # Grid defined by unique values
         low_bnd = np.percentile(x, percentiles[0])
         high_bnd = np.percentile(x, percentiles[1])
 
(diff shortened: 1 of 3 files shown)


src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -157,9 +158,11 @@ def compute_ale_1d_continuous(
     X,
     feature_idx,
     grid_resolution="auto",
+    percentiles=(0.05, 0.95),
```
bthirion:
```suggestion
    percentiles=(5, 95),
```

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -348,64 +373,87 @@ def compute_ale_1d_discrete(
...
-    local_effects_high = _predict_fn(
-        estimator, X_high[mask_high]
-    ) - _predict_fn(estimator, X_filtered[mask_high])
-
-    # Average local effects within each bin
-    combined_idx = np.concatenate(
-        [value_idx[mask_low], value_idx[mask_high] + 1]
-    )
-    combined_effects = np.concatenate([local_effects_low, local_effects_high])
+    def _compute_ale_curve(X_sample):
```
bthirion: I thought that you would have to compute the curve on fixed bins, so that the stdv makes sense ?

bthirion (commented):
Thx ! I think that we need to finalize before going to that one.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: It's what is done by calculating the unique_values outside of the function which compute the curve for a given random draw with replacement from the dataset

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -73,9 +74,11 @@ def compute_ale_1d(
     feature_type,
     method="predict",
     grid_resolution="auto",
+    percentiles=(5, 95),
     confidence_interval=True,
```
jpaillard: I think we could remove `confidence_interval=True` from the argument and simply use `if confidence_level is not None` instead. 

Wanted to comment that on #701 but was to slow^^

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -73,9 +74,11 @@ def compute_ale_1d(
...
     confidence_level=0.95,
-    percentiles=(5, 95),
+    n_bootstraps=20,
+    n_jobs=-1,
```
jpaillard: Keep 1 by default (as in scikit-learn). 
```suggestion
    n_jobs=1,
```

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -110,13 +113,17 @@ def compute_ale_1d(
...
     confidence_interval : bool, default=True
         Whether to compute the confidence intervals of the ALE curve.
     confidence_level : float, default=0.95
         The confidence level used to compute the confidence intervals (e.g., 0.95 for 95%).
-    percentiles : tuple of float, default=(5, 95)
-        The lower and upper percentile used to create the extreme values for the grid.
-        Must be in [0, 100].
+    n_bootstraps : int, default=20
+        Number of bootstrap samples to generate if `confidence_interval` is True.
+    n_jobs : int, default=-1
```
jpaillard:
```suggestion
    n_jobs : int, default=1
```

src/hidimstat/visualization/accumulated_local_effects.py
jpaillard: Would it be possible to rename this something like `X_j` to be more explicit?

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -187,76 +180,111 @@ def compute_ale_1d(
...
+        X_bootstrap = X_filtered
+
+    else:
+        raise ValueError(
+            "'feature_type' must be a string among 'continuous' and 'categorical'."
+        )
+
+    def _compute_ale_curve(X_sample):
+        """Internal closure to compute the ALE curve on a fixed grid."""
+        if feature_type == "continuous":
```
jpaillard: There's a lot of duplicated code between the if and else, this could be moved above the if statement.

src/hidimstat/visualization/accumulated_local_effects.py
jpaillard, line 20: I think this function duplicates existing code. 
To check the "method", we already have https://github.com/mind-inria/hidimstat/blob/f191c8a6952e95bdade7b642108eaf2bf11e3a1c/src/hidimstat/_utils/utils.py#L41

Then we only need to handle the predict_proba case by taking `pred[:, 1]` which may not require a dedicated function.

test/visualization/test_accumulated_local_effects.py
```diff
@@ -205,11 +206,26 @@ def test_compute_ale_1d_continuous(ale_test_data):
     assert len(grid_values_ci) <= grid_resolution + 1
     assert len(ale_err_ci) == len(grid_values_ci)
 
+    # To test _bootstrap()
+    with parallel_backend("threading"):
```
jpaillard: Why do we need to change the backend?

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -187,76 +180,111 @@ def compute_ale_1d(
...
+            raise ValueError("'n_bootstrap' must be strictly greater than 0.")
 
-        var_of_mean = np.zeros(n_bins, dtype=float)
-        valid_bins = bin_counts > 1
-        var_of_mean[valid_bins] = sum_squared_deviations[valid_bins] / (
-            bin_counts[valid_bins] * (bin_counts[valid_bins] - 1)
+        def _bootstrap():
+            bootstrap_indices = np.random.default_rng().choice(
+                len(X_bootstrap), size=len(X_bootstrap), replace=True
+            )
```
jpaillard: This will prevent the control of randomness: it will not be possible to obtain the exact same plot twice. 
What you can do is: 
 1. Use `random_state` as an argument of the function
 2. At the beginning of the function set `rng = np.random.default_rng(random_state)`
 3. bootstrap_curves = Parallel(n_jobs=n_jobs)(
            delayed(_compute_ale_curve)(X_bootstrap[rng.choice(
                len(X_bootstrap), size=len(X_bootstrap), replace=True
            )]) for _ in range(n_bootstraps)
        )

test/visualization/test_accumulated_local_effects.py
abaziree: Otherwise the function executed with Parallel isn't considered tested by pytest (probably because it is not seen in the thread which is running the tests), and this even if the number of threads is fixed to 1.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: It duplicates maximum 4 lines from the above function which only test if a given method is valid or not. Honestly I prefer to keep it like that to avoid having to copy all the checks (fitted model, Dataframe handling, multiclass) everywhere `_predict_fn` is currently used.

abaziree: yup

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: Thx I wasn't aware of the default used in scikit-learn.

test/visualization/test_accumulated_local_effects.py
jpaillard: I see. Maybe we can avoid having a function for this, then, see the comment below.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: In this case should we keep the default `confidence_level=0.95` or change it to `confidence_level=None` ?

abaziree: It's just I find that strange to have like an option which is not an integer and is not the default but that is not very important

bthirion: +1 for the suggestion of @GrituX

bthirion: the default can remain (5, 95)

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -171,92 +169,127 @@ def compute_ale_1d(
...
-        ale = np.array([0, *np.cumsum(mean_effects)])
-    else:
-        ale = np.cumsum(mean_effects)
+        return ale_curve
 
-    # Center: subtract the sample-weighted mean
-    ale_centers = (ale[1:] + ale[:-1]) / 2
-    ale -= np.sum(ale_centers * bin_counts[min_weight_bin:]) / bin_counts.sum()
+    # Base calculation
+    ale = _compute_ale_curve(X_bootstrap)
```
bthirion: I'm missing something here: in your implementation, the ale curve is computed on a different grid for each bootstrap sample, right ? Which makes it inconsistent to average across bootstraps;
the grid  should be computed one and for all and provided as argument to `_compute_ale_curve`.

abaziree: At the beginning of the function we compute the grid depending the feature's type 
https://github.com/mind-inria/hidimstat/blob/4aeff1a8cf8f2123c6eb48839237da428732e564/src/hidimstat/visualization/accumulated_local_effects.py#L146-L187
Then we can simply use `grid_values` in `_compute_ale_curve` without providing it as an argument since the variable is declared over this function.
Therefore it is the same binning which is used for each bootstrap.

bthirion: To me it's much cleaner to integrate the argument in the call of the function.
Subfunctions are not great, we may want to reuse this code in another module.

src/hidimstat/visualization/accumulated_local_effects.py
```diff
@@ -171,92 +169,127 @@ def compute_ale_1d(
...
-        mask_low = x_filtered != grid_values[0]
-        mask_high = x_filtered != grid_values[-1]
-        X_low[mask_low, feature_idx] = grid_values[value_idx[mask_low] - 1]
-        X_high[mask_high, feature_idx] = grid_values[value_idx[mask_high] + 1]
+    else:
+        raise ValueError(
+            "'feature_type' must be a string among 'continuous' and 'categorical'."
+        )
+
+    def _compute_ale_curve(X_sample):
```
GrituX: Please avoid nested functions

src/hidimstat/visualization/accumulated_local_effects.py
GrituX: I think there is some redundancy in the arguments. Instead of being `None`, `confidence_level` could be defaulted to 0, and the check could be on that. Also, to hint that it's an option, can a rename as `with_confidence_level` be helpful ? It's a bit long though.

src/hidimstat/visualization/accumulated_local_effects.py
abaziree: okay

bthirion (approved):
LGTM, thx !

Issue #680, opened 2026-06-17
[DOC] Add example using TabICL


GrituX:
### Please confirm that you have searched existing issues in the repository.

- [x] Yes

### Issue with current documentation:

No existing example of a tabular foundation model.

### Idea or request for content:

For the documentation rework, implement an example with the Adult dataset for instance and TabICL.

bthirion:
Sounds good.

PR #704, opened 2026-07-02
[DOC] Add PyTorch / Skorch example

GrituX:
I added an example to compute feature importance on pixels for binary classification on the MNIST digit dataset with a CNN estimator written in PyTorch and that interfaces with Scikit-Learn through the Skorch library.
The version limitation of Torch and Skorch come from Skorch as the support of `__sklearn_is_fitted__` for estimators only arrived on 1.3 and 1.3.1 adds the support of sklearn v1.8 and 1.9.


examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,245 @@
+"""
+Using PyTorch with HiDimStat
+============================
+HiDimStat was designed with the goal of being easily compatible with Sklearn.
+PyTorch models might not be used directly with HiDimStat for that reason.
+However, with the help of a third party library `Skorch <https://skorch.readthedocs.io/en/stable/>`,
+PyTorch can be interfaced with HiDimStat, and provide all of its functionalities.
+In this example, we show how to define a Convolutional Neural Network (CNN) in Skorch,
```
bthirion:
```suggestion
In this example, we  define a Convolutional Neural Network (CNN) in Skorch,
```

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,245 @@
...
+and perform pixel-wise feature importance in a binary classification setup,
+through the MNIST digits dataset, between digits 4 vs 7 and 0 vs 1.
+"""
+
+# %%
+# Deefining a PyTorch model
```
bthirion:
```suggestion
# Defining a PyTorch model
```

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,245 @@
...
+# -------------------------
+# We start by defining a basic Convolutional Neural Network (CNN)
+# composed of 2 series of (Convolution, Activation, Pooling) layers,
+# followed by 2 fully-connected layers.
+
+import torch
```
bthirion: needed ?

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,245 @@
...
+)
+model = make_pipeline(StandardScaler(), net)
+# Current hotfix of API compatibility issue (Skorch doesn't automatically set this during fitting).
+# This needs to be set, otherwise HiDimStat throws an error.
+net.n_features_in_ = 28 * 28
+
+# %%
+# Running HiDimStat feature importance computation
+# ------------------------------------------------
+# We cluster pixels through feature agglomeration, and define its connectivity
```
bthirion:
```suggestion
# We cluster pixels through feature agglomeration, while leveraging their grid structure
```

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,245 @@
...
+# and the values are the list of column names corresponding to each features group.
+order = np.argsort(clustering.labels_)
+sorted_ids = clustering.labels_[order]
+unique_ids, start_idx = np.unique(sorted_ids, return_index=True)
+positions = np.split(order, start_idx[1:])
+features_groups = dict(zip(unique_ids, positions, strict=False))
+# features_groups = {idx: [cluster_label] for idx, cluster_label in enumerate(clustering.labels_)}
+
+# Careful when using Skorch, having n_jobs > 1 might create joblib and pickle issues.
+pfi = PFI(
```
bthirion: Why not CFI  instead ?

bthirion (commented):
Thx ! Please find a few comments enclosed

examples/plot_advanced_skorch.py
GrituX: CFI works too, it's just slower than PFI and requires more samples (2000 against 500) otherwise the solver often stumbles on situations where only 1 class is present in the data. But the runtime is still fine, so I made the switch.

GrituX:
Since this PR doesn't target the `main` branch, the CI will never be run ... So to see the result, this would need to be merged into the target branch, which this time targets `main` ...

bthirion:
Where can one see the generated doc ?

test/test_leave_one_covariate_in.py
```diff
@@ -0,0 +1,253 @@
+from functools import partial
```
bthirion, line 1: this is unrelated, right ?

src/hidimstat/leave_one_covariate_in.py
```diff
@@ -0,0 +1,361 @@
+import numpy as np
```
bthirion, line 1: this is unrelated, right ?

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,247 @@
...
+    fwer=target_fwer, n_tests=n_clusters, two_tailed_test=True
+)
+
+# %%
+# Visualizing the results
+# -----------------------
+# Finally, we visualize the significant pixels identified by CFI for each of the
+# classification tasks.
+
+import matplotlib.pyplot as plt
```
bthirion: already imported

bthirion (commented):
LGTM overall, just beware LOCI code that seems unrelated + I'd need to see the generatde doc: how does it render ? Can/should we shorted it a little bit ?

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,247 @@
...
+mask_0_1 = (y_mnist == "0") | (y_mnist == "1")
+X_0_1, y_0_1 = X_mnist[mask_0_1], y_mnist[mask_0_1].astype(int)
+X_0_1, y_0_1 = resample(
+    X_0_1,
+    y_0_1,
+    n_samples=n_samples,
+    replace=False,
+    random_state=0,
+    stratify=y_0_1,
+)
```
jpaillard: The example could be streamlined to focus only on `skorch`; we could remove the second classification task.

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,247 @@
...
+sorted_ids = clustering.labels_[order]
+unique_ids, start_idx = np.unique(sorted_ids, return_index=True)
+positions = np.split(order, start_idx[1:])
+features_groups = dict(zip(unique_ids, positions, strict=False))
+
+# %%
+# We now instantiate the CFI and fit it on the two tasks.
+
+# Careful when using Skorch, having n_jobs > 1 might create joblib and pickle issues.
+cfi = CFI(
```
jpaillard, line 157: If we use CFI, we can also plot the importance rather than the selection.

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,247 @@
...
+
+import torch.cuda
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from skorch import NeuralNetClassifier
+from torch.optim import Adam
+
+net = NeuralNetClassifier(
+    MNISTCNN,
+    max_epochs=5,
```
jpaillard: This looks small. Also, I would suggest reporting the prediction performance on a held-out test set. To emphasize that being model-agnostic allows the use of powerful predictors.

examples/plot_advanced_skorch.py
```diff
@@ -0,0 +1,247 @@
...
+
+for i, (title, selected) in enumerate(
+    [
+        ("4 vs 7", selected_4_7),
+        ("0 vs 1", selected_0_1),
+    ]
+):
+    pixel_selection = np.zeros((len(clustering.labels_),), dtype=int)
+    # selected contains the selected cluster groups, so we convert this back to pixel selection.
+    for sign in (-1, 1):
```
jpaillard: For CFI, there should be no negative labels on the selection map. This is only for methods based on the desparsified Lasso, where the Lasso coefficient signs indicate the direction of the effect.

src/hidimstat/leave_one_covariate_in.py
GrituX: This is weird, the imports were already there. I don't know why these are considered changes.

examples/plot_advanced_skorch.py
GrituX: I added the importance to the selection for comparison

examples/plot_advanced_skorch.py
GrituX: Also the method is slightly convoluted. I realized that clustering has an `inverse_transform` method which replaces all this code.

GrituX:
I applied changes according to reviews. Unfortunately, the current CI doesn't build documentation for PRs that do not target main. So I suggest that once code review is over, we merge this into the target branch, and discuss the generated doc there.

GrituX:
There is the same problem in this PR as with the TabICL in #680 where LOCI is included. I will close this PR and open a new one that directly targets main, so that we have doc building. It's easier to discuss from there.

PR #705, opened 2026-07-02
[DOC] TabICL tabular foundation model

GrituX:
I added a basic example of TabICL usage with HiDimStat on the California Housing dataset. Unfortunately, Github doesn't give free access to GPUs, so I limited the example to 1000 samples which still amounts to a runtime of 1min30s for the entire script.
Also, TabICL currently runs only with CFI and PFI. TabICL requires an integer seeding, while D0CRT directly works with a seeding generator object, making them conflict.
The example is basic, I did not manage to find something interesting to do with it that did not explode computation time since everything has to run on CPU.


examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,82 @@
+"""
+Tabular Foundation Model TabICL
+================================
+
+In this example, we demonstrate how to use a tabular foundation model such as
+TabICL [:footcite:t:`qu2025tabicl`] with a straightforward example on the California
+Housing regression dataset.
+"""
+
+import pandas as pd
```
bthirion: please perform the import wherever they're needed. It's indeed not standard, but makes the example more readable.

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,82 @@
...
+X, y = resample(
+    X, y, replace=False, n_samples=1000, stratify=y, random_state=0
+)
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.2, random_state=0
+)
+
+model = make_pipeline(StandardScaler(), TabICLRegressor(device="cpu"))
+model.fit(X_train, y_train)
+
```
bthirion, line 41: maybe report model accuracy on the test data ?

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,82 @@
...
+)
+sns.despine()
+plt.show()
+
+# %%
+# As you can see from the timer under, even with 1000 samples and 9 features,
+# running TabICL on the CPU is slow. We recommend to use it on larger datasets,
+# be it in terms of samples and/or features and to run it on GPU for faster
+# inferences. Due to technical limitations, this is something we cannot showcase
+# here.
```
bthirion: you can skip the last sentence imho.
The example LGTM.

GrituX:
I made the changes following your reviews. It's not the most interesting example, but it shows that we can interface HiDimStat with SOTA models.

examples/plot_advanced_tabicl.py
GrituX: Model accuracy is of 0.743

GrituX:
However, I can't seem to understand why the CI doesn't want to build the documentation even with the [doc] tag that forces the doc build.

GrituX:
After a look at the CI, since this PR doesn't target the main branch, the CI will never be run ... So to see the result, this would need to be merged into the target branch, which this time targets main ... (same issue as with the PyTorch/Skorch example)

examples/plot_loci.py
```diff
@@ -0,0 +1,162 @@
+"""
```
bthirion, line 1: this is unrelated

bthirion (commented):
The tabular part looks good, but the PR also includes the LOCI PR, which it should not.
Maybe open a new one ?

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,87 @@
...
+ax = sns.barplot(
+    data=df,
+    y="feature",
+    x="importance",
+    hue="selected",
+    palette="muted",
+    orient="h",
+)
+sns.despine()
+plt.show()
```
jpaillard, line 77: Here we can use `cfi.plot()` which calls `sns.barplot`

jpaillard (commented):
I guess LOCO/LOCI are also supported. 

Note: for "in context learning" models, the fit is very cheap (only cache a few things) and the predict expensive (compute the actual attention operations). So actually for such models, a naive implementation makes LOCO (based on many fit calls) cheaper than CFI (based on many predict calls). 
This can be mitigated by using KV-caching (computing only once the attention operation using $X_{train}, y_{train}$ and caching it). TabICL and TabPFN provide wrappers for that. 

I'll try to suggest a code snippet for this.

jpaillard:
Do you know why the doc is not being built?

GrituX:
The doc is not built for the same reason as with the PyTorch example : this PR does not target the main branch. So the CI doesn't build the doc.

GrituX:
@bthirion for the LOCI bit, I don't know why this happens. It's not the first PR to have this issue. I can open a new one that also directly targets the main so that documentation is built.

GrituX:
Discussion should now continue in #716

Issue #562, opened 2025-12-10
Use of DepsarsifiedLasso in CluDL is tricky


bthirion:
using default parameters when creating CluDL yields weird results, as can be seen by running
```
import numpy as np
from sklearn.feature_extraction import image
from sklearn.cluster import FeatureAgglomeration
from sklearn.linear_model import LassoCV
from hidimstat.desparsified_lasso import DesparsifiedLasso
from hidimstat._utils.scenario import multivariate_simulation_spatial
from hidimstat import CluDL

n_samples = 100
shape = (40, 40)
roi_size = 4  # size of the edge of the four predictive regions
X_init, y, beta, epsilon = multivariate_simulation_spatial(
    n_samples, shape, roi_size, signal_noise_ratio=10., smooth_X=1)

alpha = .05 # alpha is the significance level for the statistical test
n_clusters = 200
connectivity = image.grid_to_graph(n_x=shape[0], n_y=shape[1])
ward = FeatureAgglomeration(
    n_clusters=n_clusters, connectivity=connectivity, linkage="ward")

dl = DesparsifiedLasso(LassoCV())

c1 = CluDL(clustering=ward)
c1.fit_importance(X_init, y)
selected_cdl = c1.fwer_selection(alpha, n_tests=n_clusters)
print(f'Clustered Desparsified Lasso selected {np.sum(selected_cdl)} features among {np.sum(beta > 0)} ')

c2 = CluDL(clustering=ward, cluster_boostrap_size=.9999)
c2.fit_importance(X_init, y)
selected_cdl = c2.fwer_selection(alpha, n_tests=n_clusters)
print(f'Clustered Desparsified Lasso selected {np.sum(selected_cdl)} features among {np.sum(beta > 0)} ')

import matplotlib.pyplot as plt
plt.figure(figsize=(10, 6))
ax = plt.subplot(121)
ax.matshow(c1.importances_.reshape(shape))
ax = plt.subplot(122)
ax.matshow(c2.importances_.reshape(shape))
plt.show()
```

By contrast, replacing the line
```
c2 = CluDL(clustering=ward, cluster_boostrap_size=.9999)
```
by
```
c2 = CluDL(clustering=ward, cluster_boostrap_size=.9999, desparsified_lasso=dl)
```
works.

The reason seems to be that with the default call, the same object is reused... with refitting. It is unclear to me why.
[...]

PR #707, opened 2026-07-06
[FIX] assert fix on CluDL independence test

### LINKED ISSUE #706: [BUG] False assert in test_cludl_independance

- [x] Yes

### Describe your issue.

An assertion can be false when running test_cludl_independance. It seems to be in a particular situation since it is the first time I encounter this problem when running tests.

Image

### Reproducing Code Example

```python
def test_cludl_independence():
    """Test that CluDL works with repeated calls
    non-regression test for #425
    """
    n_samples = 50
    shape = (20, 20)
    roi_size = 4  # size of the edge of the four predictive regions
    X_init, y, _, _ = multivariate_simulation_spatial(
        n_samples, shape, roi_size, signal_noise_ratio=10.0, smooth_X=1
    )
    alpha = 0.05  # alpha is the significance level for the statistical test
    n_clusters = 50
    connectivity = image.grid_to_graph(n_x=shape[0], n_y=shape[1])
    ward = FeatureAgglomeration(
        n_clusters=n_clusters, connectivity=connectivity, linkage="ward"
    )

    c1 = CluDL(clustering=ward)
    c1.fit_importance(X_init, y)
    s1 = c1.fwer_selection(alpha, n_tests=n_clusters)
    c2 = CluDL(clustering=ward, cluster_bootstrap_size=0.5)
    c2.fit_importance(X_init, y)
    s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
    assert np.sum(s2) > np.sum(s1 / 2)
```

### Error message

```shell
FAILED test/test_ensemble_clustered_inference.py::test_cludl_independence - asser np.int64(32) > np.float64(32.0)
```

```shell
0.3.2.dev23+g2a14fcb65.d20260625 sys.version_info(major=3, minor=12, micro=3, releaselevel='final', serial=0)
```

GrituX:
`test_cludl_independence` sometimes failed its final assert due to variance issues. The fix is to repeat the procedure and check that it passes the assertion in average over 5 iterations.

### DIFF
--- a/test/test_ensemble_clustered_inference.py
+++ b/test/test_ensemble_clustered_inference.py
@@ -127,13 +127,18 @@ def test_cludl_independence():
     c1.fit_importance(X_init, y)
     s1 = c1.fwer_selection(alpha, n_tests=n_clusters)
     s2_iterations = np.zeros((0, len(s1)))
-    nb_iterations = 5
+    nb_iterations = 20
     for _ in range(nb_iterations):
         c2 = CluDL(clustering=ward, cluster_bootstrap_size=0.5)
         c2.fit_importance(X_init, y)
         s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
         s2_iterations = np.vstack((s2_iterations, s2))
-    assert np.sum(s2_iterations) / nb_iterations > np.sum(s1) / 2
+
+    assert np.sum(s1) != 0
+    assert (
+        np.abs(np.sum(s2_iterations) / nb_iterations - np.sum(s1)) / np.sum(s1)
+        < 0.2
+    )
 
 
 def test_encludl_spatial():
--- a/test/test_ensemble_clustered_inference.py
+++ b/test/test_ensemble_clustered_inference.py
@@ -127,8 +127,8 @@ def test_cludl_independence():
     c1.fit_importance(X_init, y)
     s1 = c1.fwer_selection(alpha, n_tests=n_clusters)
     s2_iterations = np.zeros((0, len(s1)))
-    nb_iterations = 20
-    for _ in range(nb_iterations):
+    n_iterations = 20
+    for _ in range(n_iterations):
         c2 = CluDL(clustering=ward, cluster_bootstrap_size=0.5)
         c2.fit_importance(X_init, y)
         s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
@@ -136,8 +136,8 @@ def test_cludl_independence():
 
     assert np.sum(s1) != 0
     assert (
-        np.abs(np.sum(s2_iterations) / nb_iterations - np.sum(s1)) / np.sum(s1)
-        < 0.2
+        np.abs(np.sum(s2_iterations) / n_iterations - np.sum(s1)) / np.sum(s1)
+        < 0.5
     )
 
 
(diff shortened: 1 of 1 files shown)


jpaillard (commented):
I'm not sure what is tested here (before the modification introduced in this PR). 

It was introduced in #562 after we discovered a bug when using the default value of `estimator` in `CluDL`. Of note, this was then fixed when we stopped using sklearn estimators as default values for arguments. 

I think @bthirion wanted to check that CluDL still "works" when we repeat `fit_importance` with two different instanecs using different `cluster_bootstrap_size`. However, I don't see why we should have `np.sum(s2) > np.sum(s1 / 2)` or `np.sum(s2) > np.sum(s1) / 2`. 
Maybe we simply want to check something like ` (np.sum(s1) > 0) & (np.sum(s2) > 0)`

bthirion:
The rationale for  `np.sum(s2) > np.sum(s1 / 2)` or `np.sum(s2) > np.sum(s1) / 2` is that the number of detections should be approximately similar for different draws.

jpaillard:
I'm not sure I understand the test then. Isn't it testing that the number of discoveries is twice as large for the second draw?

bthirion:
> I'm not sure I understand the test then. Isn't it testing that the number of discoveries is twice as large for the second draw?

IIUC it's testing that it's "more than half" i.e. roughly equal.

GrituX:
If we're testing rough equality, wouldn't it be better to have something such as `abs(np.sum(s2) - np.sum(s1)) / np.sum(s1) < threshold` where threshold is something like 0.02 ?

bthirion:
> If we're testing rough equality, wouldn't it be better to have something such as `abs(np.sum(s2) - np.sum(s1)) / np.sum(s1) < threshold` where threshold is something like 0.02 ?

Sure. just beware that these numbers are nonzero, and do not choose an overly tight threshold.

test/test_ensemble_clustered_inference.py
```diff
@@ -126,10 +126,19 @@ def test_cludl_independence():
     c1 = CluDL(clustering=ward)
     c1.fit_importance(X_init, y)
     s1 = c1.fwer_selection(alpha, n_tests=n_clusters)
-    c2 = CluDL(clustering=ward, cluster_bootstrap_size=0.5)
-    c2.fit_importance(X_init, y)
-    s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
-    assert np.sum(s2) > np.sum(s1 / 2)
+    s2_iterations = np.zeros((0, len(s1)))
+    nb_iterations = 20
```
bthirion:
```suggestion
    n_iterations = 20
```

test/test_ensemble_clustered_inference.py
```diff
@@ -126,10 +126,19 @@ def test_cludl_independence():
...
+    for _ in range(nb_iterations):
```
bthirion:
```suggestion
    for _ in range(n_iterations):
```

test/test_ensemble_clustered_inference.py
```diff
@@ -126,10 +126,19 @@ def test_cludl_independence():
...
+        c2 = CluDL(clustering=ward, cluster_bootstrap_size=0.5)
+        c2.fit_importance(X_init, y)
+        s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
+        s2_iterations = np.vstack((s2_iterations, s2))
+
+    assert np.sum(s1) != 0
+    assert (
+        np.abs(np.sum(s2_iterations) / nb_iterations - np.sum(s1)) / np.sum(s1)
+        < 0.2
```
bthirion: Why not relax it to .5 ? To avoid random failures.

bthirion (commented):
Thx. I only have some minor comments.

jpaillard (approved):

PR #710, opened 2026-07-07
[ENH] Parallelize importance in EnCluDL

### LINKED ISSUE #709: [ENH] Parallelize importance computation in EnCluDL
Right now we have in EnCluDL:
```python
for i in tqdm(
            range(self.n_bootstraps),
            desc="Computing importances",
            total=self.n_bootstraps,
        ):
            self.clustering_desparsified_lassos_[i].importance()
```
Instead we could wrap the code with joblib.

GrituX:
Importance was computed sequentially for each bootstrap in EnCluDL, so I parallelized it with joblib.

### DIFF
--- a/src/hidimstat/ensemble_clustered_inference.py
+++ b/src/hidimstat/ensemble_clustered_inference.py
@@ -404,6 +404,11 @@ class EnCluDL(BaseVariableImportance):
 
         return self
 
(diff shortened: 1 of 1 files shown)


jpaillard (approved):
Thx!

PR #715, opened 2026-07-09
[FEAT] Added generic Ensemble Clustered Variable Importance

### LINKED ISSUE #714: [ENH] Ensemble Clustered Generic VI Estimator

- [x] Yes

### Please describe the feature that you would like to see improved or added.

We talked about this briefly once with @bthirion. I looked at the code, and I actually think that this could be very easily done. With very few line code changes, the current EnCluDL could easily be turned into a class that encapsulates any variable importance estimator. If we include this, the execution time for some more time-consuming estimators should be checked beforehand to provide a few guidelines and what to expect in terms of compute performances.

### Describe the solution you would like to see implemented.

_No response_

### Describe alternatives you have considered.

_No response_

_No response_

GrituX:
Just as an example for the moment, but this came with little changes from EnCluDL.
For the moment, the tests for the method only run with a desparsified lasso. I still have to check the performances of the method with other variable importance methods.


GrituX:
Code now fully runs on most vi methods. Model X Knockoffs doesn't work for the moment, as importances are nested inside a list of length n_repetitions.
I added an [example](https://output.circle-artifacts.com/output/job/7c9d66e1-cb7d-4c50-815e-09de72509d1d/artifacts/0/documentation/generated/gallery/examples/plot_multi_encluvi.html) to show that it can run on all other methods (with an execution timer). The example is currently designed for discussion, and is very simple, we can maybe find a more suitable example to showcase the benefits of EnCluVI.

bthirion:
Do you want a review ?
It's OK if does not work (yet) with KOs.

GrituX:
Yes please ! Other methods work, so the code is ready for reviews.

examples/plot_multi_encluvi.py
```diff
@@ -0,0 +1,175 @@
...
+        method="predict_proba",
+        loss=log_loss,
+    ),
+    LOCI(
+        estimator=clf,
+        method="predict_proba",
+        loss=log_loss,
+    ),
+]
+
```
bthirion: I think that having all these performance estimators clutters the example and makes it hard to understand.
I would narrow down to CPI and LOCO, or maybe CPI and a linear estimators (e.g. dCRT), but no more than that.
(PFI and LOCI are not to be recommended, for different reasons, so we should not use them in many examples)

examples/plot_multi_encluvi.py
```diff
@@ -0,0 +1,175 @@
...
+n_cols = 3
+n_rows = (nb_vis - 1) // n_cols + 1
+
+_, axes = plt.subplots(nrows=n_rows, ncols=n_cols, figsize=(10, 3 * n_rows))
+for i in range(nb_vis):
+    ax = vis[i].plot_importance(ax=axes[i // n_cols, i % n_cols])
+    ax.set_xlabel(f"{type(vis[i]).__name__}")
+plt.tight_layout()
+plt.show()
+
```
bthirion: If the features have a more meaningful name, that would be nicer...

src/hidimstat/ensemble_clustered_variable_importance.py
```diff
@@ -0,0 +1,518 @@
...
+        if self.vi_estimator_ is None:
+            raise ValueError(
+                "The estimator requires to be fit before to use them"
+            )
+        check_is_fitted(self.vi_estimator_)
+
+
+class EnCluVI(BaseVariableImportance):
+    """
+    Ensemble clustered inference with desparsified lasso. Performs multiple runs
```
bthirion:
```suggestion
    Ensemble clustered inference with arbitrary variable importance measure.
```

bthirion (commented):
IIUC EnCluDL becomes as special case of EnCluVI, right ?
In that case, I would deprecate it (e.g. to be removed in 0.5.0).
Can be in a forthcoming PR if need be

GrituX:
Sorry, I forgot to mention that the example plot_multi_encluvi was designed just for this PR to show that the method works on all VIM but Knockoffs. I will delete it before merging once this PR is over.

GrituX:
Following @jpaillard suggestion, I changed the class names and signatures, moved each class to a separate file to make EnsembleImportance more generic.
Since there was a sampling utility function that was used by both classes, I created a utils file in the "samplers" module. Lmk if you think it's unnecessary, in which case I will find another spot for this method.

src/hidimstat/ensemble_clustered_inference.py
```diff
@@ -52,6 +54,11 @@ class CluDL(BaseVariableImportance):
     n_features_ : int
         Number of features in the original data.
 
+    Notes
+    -----
+    .. deprecated:: 0.5.0
+       Use :class:`hidimstat.ClusterImportance` instead.
```
bthirion: with the DesparsifiedLasso estimator.

bthirion:
Thank you. Can we clarify what other ensembles are intended, on top on EncluVI ?

src/hidimstat/samplers/utils.py
jpaillard, line 1: This duplicates https://github.com/mind-inria/hidimstat/blob/736b2166386b276ba425c0eb16f089317ae418da/src/hidimstat/ensemble_clustered_inference.py#L237

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Parameters
+    ----------
+    vim: hidimstat.BaseVariableImportance
+        Any variable importance method that derives from hidimstat's BaseVariableImportance class.
+    n_repeats: int, optional (default=25)
+        Number of bootstrap iterations for ensemble inference.
+    bootstrap_frac: float, optional (default=0.3)
+        Fraction of samples used for the ensemble.
+        When bootstrap_frac=1.0, all samples are used.
```
jpaillard, line 28: It would make sense when `bootstrap_frac=1` to still make bootstrap draws, with replacement. Something like `resample(replace=True, ...)`, which is not currently the case

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def __init__(
+        self,
+        vim,
+        n_repeats=25,
+        bootstrap_frac=0.3,
+        bootstrap_groups=None,
+        n_jobs=1,
+        random_state=None,
+        memory=None,
```
jpaillard:
```suggestion
```

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,238 @@
...
+        gamma=0.5,
+        adaptive_aggregation=False,
+    ):
+        self.vim = vim
+        self.n_repeats = n_repeats
+        self.bootstrap_frac = bootstrap_frac
+        self.bootstrap_groups = bootstrap_groups
+        self.n_jobs = n_jobs
+        self.random_state = random_state
+        self.memory = memory
```
jpaillard:
```suggestion
```

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,238 @@
...
+        Fraction of samples used for the ensemble.
+        When bootstrap_frac=1.0, all samples are used.
+    bootstrap_groups: ndarray, shape (n_samples,), optional (default=None)
+        Sample group labels for stratified subsampling.
+    n_jobs : int or None, optional (default=1)
+        Number of parallel jobs.
+    random_state: int, optional (default=None)
+        Random seed for reproducible subsampling.
+    memory : joblib.Memory or str, optional (default=None)
+        Used to cache the output of the clustering and inference computation.
```
jpaillard: Not used
```suggestion
```

src/hidimstat/leave_one_covariate_out.py
jpaillard, line 1: Is this meant to be in this PR?

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.n_features_in_ = X.shape[1]
+
+        # Clustering
+        self.clustering_samples_ = _subsampling(
+            n_samples=X.shape[0],
+            train_size=self.cluster_frac,
+            groups=self.bootstrap_groups,
+            random_state=rng,
+        )
```
jpaillard: I would put the subsampling only in `EnsembleImportance` and remove it from `ClusterImportance` for clarity. That way the API would be easier to understand: 
 - `ClusterImportance` takes X, y, and returns importance per cluster
 - `EnsembleImportance` takes X, y, generates sub-samples, and computes importance for each sub-sample. 

With this approach, a user could still get clustered importance on sub-samples by calling `EnsembleImportance(vim=ClusterImportance())`

jpaillard (commented):
For ModelX-KO, it currently already contains the "ensemble" part with the `n_repeats` parameter. Maybe we can unify its API in a next PR by simply removing the `n_repeats` which would now be obtained with `EnsembleImportance(vim=ModelXKO)`

src/hidimstat/ensemble_importance.py
GrituX: Indeed. I'll make the change.

src/hidimstat/leave_one_covariate_out.py
GrituX: Idk why this is there. I'm reverting the changes.

src/hidimstat/cluster_importance.py
GrituX: I agree it makes more sense.

GrituX:
Since we're getting close to the end of this PR, I removed the demo example `plot_multi_encluvi.py` and made the changes from EnCluDL to EnsembleImportance in `plot_2D_simulation_example.py`.

src/hidimstat/ensemble_importance.py
bthirion: Agree, but I would do that in another PR.

test/test_cluster_importance.py
```diff
@@ -0,0 +1,234 @@
...
+    c1 = ClusterImportance(
+        vim=vim,
+        clustering=ward,
+    )
+    c1.fit_importance(X_init, y)
+    s1 = c1.fwer_selection(alpha, n_tests=n_clusters)
+    c2 = ClusterImportance(vim=vim, clustering=ward, cluster_frac=0.5)
+    c2.fit_importance(X_init, y)
+    s2 = c2.fwer_selection(alpha, n_tests=n_clusters)
+    assert np.sum(s2) > np.sum(s1 / 2)
```
bthirion:
```suggestion
    assert np.sum(s2) > np.sum(s1) / 2
```

docs/src/api.rst
jpaillard, line 1: Maybe it would make sense to have a section for this kind of "meta importane methods" like `ClusterImportance`, `EnsembleImportance`, `CluDL` ... 
I am not so inspired but it could be something like "Grouping & Aggregation" or "Importance pipelines"

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+
+    def __init__(
+        self,
+        vim,
+        clustering,
+        cluster_frac=1.0,
+        bootstrap_groups=None,
+        random_state=None,
+        memory=None,
```
jpaillard: This is not used

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,226 @@
...
+            fdr_control=fdr_control,
+            reshaping_function=reshaping_function,
+            two_tailed_test=two_tailed_test,
+        )
+
+    @staticmethod
+    def _ungroup_importance(importance, n_features, ward):
+        """
+        Ungroup cluster-level beta coefficients to individual feature-level
+        coefficients.
```
jpaillard:
```suggestion
        Ungroup cluster-level importances to individual feature-level
        importances.
```

test/test_ensemble_importance.py
jpaillard, line 1: Could we also have a test for the reproducibility? 
See https://hidimstat.github.io/dev/dev/developer_guidelines.html#testing

docs/src/api.rst
GrituX: Yes, it doesn't really belong with other feature importance classes.
I think that "Grouping & Aggregation" is good

test/test_ensemble_importance.py
GrituX: Yes, I'm adding other tests that are missing too.

GrituX:
I've included more tests, some basic, and in compliance with rng testing as reminded by @jpaillard . Should be better now

examples/plot_2D_simulation_example.py
```diff
@@ -327,6 +326,7 @@ def compute_spatially_relaxed_mask(mask, beta_extended):
 ]
 axes[0].legend(handles=legend_handles, loc="lower center")
 plt.tight_layout()
+plt.show()
 
 
 # %%
```
bthirion, line 332: Can you also update the fMRI example ?

test/test_cluster_importance.py
```diff
@@ -0,0 +1,300 @@
...
+    ).flatten()
+    false_positive = np.setdiff1d(selected_ids, ground_truth_extended_ids)
+
+    fdp = len(false_positive) / len(selected_ids)
+    power = len(true_positive) / len(beta_ids)
+    return fdp, power
+
+
+def test_cluster_parameter_check():
+    with pytest.raises(
```
bthirion, line 59: can you specify what is tested here ?

test/test_cluster_importance.py
```diff
@@ -0,0 +1,300 @@
...
+        match=r"clustering needs to be an instance of sklearn\.cluster\.FeatureAgglomeration",
+    ):
+        ClusterImportance(
+            vim=DesparsifiedLasso(estimator=LassoCV()),
+            clustering=LassoCV(),
+        )
+
+
+def test_cluster_importance_check_fit():
+
```
bthirion: Same here

test/test_ensemble_importance.py
```diff
@@ -0,0 +1,411 @@
...
+    ).flatten()
+    false_positive = np.setdiff1d(selected_ids, ground_truth_extended_ids)
+
+    fdp = len(false_positive) / len(selected_ids)
+    power = len(true_positive) / len(beta_ids)
+    return fdp, power
+
+
+def test_ensemble_parameter_check():
+    with pytest.raises(
```
bthirion, line 56: missing docstring

test/test_ensemble_importance.py
```diff
@@ -0,0 +1,411 @@
...
+        AssertionError,
+        match="estimator needs to be a subclass of BaseVariableImportance",
+    ):
+        EnsembleImportance(vim=LassoCV())
+
+
+def test_ensemble_importance_check_fit():
+
```
bthirion: missing docstring

bthirion (commented):
I have relatively minor comments left, besides the fMRI data example.

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,236 @@
...
+        self.random_state = random_state
+        self.ensembling_method = ensembling_method
+        self.gamma = gamma
+        self.adaptive_aggregation = adaptive_aggregation
+
+        self.vim_ = None
+        self.ensemble_vims_ = None
+
+    @staticmethod
+    def _joblib_fit_one(
```
jpaillard, line 85: Should we consider setting the `vim.random_state` to `EnsembleImportance.random_state`. That way the user would only need to pass one `random_state`? 

Currently, if you do something like `EnsembleImportance(vim=CFI(random_state=None), random_state=42),` it's not so clear what comes out in terms of reproducibility.

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,201 @@
...
+        self,
+        vim,
+        clustering,
+    ):
+        assert issubclass(vim.__class__, BaseVariableImportance), (
+            "estimator needs to be a subclass of BaseVariableImportance"
+        )
+        assert issubclass(clustering.__class__, FeatureAgglomeration), (
+            "clustering needs to be an instance of sklearn.cluster.FeatureAgglomeration"
+        )
```
jpaillard: This should be moved to fit according to scikit-learn's convention.

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,201 @@
...
+    ):
+        """
+        Overrides the signature to set two_tailed_test=True by default.
+        """
+        return super().fdr_selection(
+            fdr=fdr,
+            fdr_control=fdr_control,
+            reshaping_function=reshaping_function,
+            two_tailed_test=two_tailed_test,
+        )
```
jpaillard: I think this was carried over from CluDL where `two_tailed_test` makes sense since we're looking at the signed coefficient of a Lasso model. If we want to keep ClusterImportance generic, this should not be forced.

jpaillard: would there be a way to call `vim.fdr_selection() ` so that it inherits the fdr method from the chosen vim?

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __init__(
+        self,
+        vim,
+        n_repeats=25,
+        bootstrap_frac=0.3,
+        bootstrap_groups=None,
+        n_jobs=1,
+        random_state=None,
+        ensembling_method="quantiles",
```
jpaillard: Maybe we could also take the opportunity to rename this. I find `ensembling` is a little confusing in this context. 
```suggestion
        aggregation="quantiles",
```

src/hidimstat/ensemble_importance.py
GrituX: I initially did this, but then deleted it to give more control over what is being controled. Since EnsembleImportance is a "meta learner", it would make sense that its `random_state` is imposed to the underlying `vim`.

src/hidimstat/cluster_importance.py
GrituX: Ok

src/hidimstat/cluster_importance.py
GrituX: Do you mean that when `clu_vi.fdr_selection()` is called, under the hood it's the default parameters from `clu_vi.vim.fdr_selection()` that are used ?

GrituX: This is doable. However, we would need to choose whether a change to one of the default parameters of `ClusterImportance.fdr_selection()` overrides this, or if we stick to the fdr method of the chosen vim no matter what.

GrituX: I suggested something in the following push. But I don't like how it closes options, and also changes the signature (since other parameters aren't used).

jpaillard: I think just `clu_vi.vim.fdr_selection()` should be fine. We already have different defaults depending on the methods. For instance DesparsifiedLasso has `two_tailed_test=True` whereas CFI has `two_tailed_test=False`

GrituX: the underlying vim fdr selection needs to be inversely transformed by the clustering, which adds a slight overhead compared to what I suggested

bthirion: > I think just `clu_vi.vim.fdr_selection()` should be fine. We already have different defaults depending on the methods. For instance DesparsifiedLasso has `two_tailed_test=True` whereas CFI has `two_tailed_test=False`

But this makes sense: significant DL can be positive or negative, while for CPI, only positive values truly matter.

GrituX: Ok ! Well the change is already in this push, so the code's ready for that.

src/hidimstat/cluster_importance.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        Overrides the default signature to use the default signature of the underlying vim.
+        """
+        vim_sig = inspect.signature(self.vim_.fdr_selection)
+        # get the default values of the fdr_selection method of the chosen vim.
+        vim_default_args = [
+            p.default
+            for p in vim_sig.parameters.values()
+            if p.default is not inspect.Parameter.empty
+        ]
```
jpaillard: One thing I'm skeptical about with this solution is that, IIUC, the user cannot overwrite the defaults. 

Like if I want the vim to be DesparsifiedLasso but would like to make the selection using a one-sided test, I cannot change this here.

I think it would make sense to allow this, which we already support for the basic (not clustered) `DesparsifiedLasso`

src/hidimstat/ensemble_importance.py
```diff
@@ -0,0 +1,239 @@
...
+        Number of bootstrap iterations for ensemble inference.
+    bootstrap_frac: float, optional (default=0.3)
+        Fraction of samples used for the ensemble.
+        When bootstrap_frac=1.0, all samples are used.
+    bootstrap_groups: ndarray, shape (n_samples,), optional (default=None)
+        Sample group labels for stratified subsampling.
+    n_jobs : int or None, optional (default=1)
+        Number of parallel jobs.
+    random_state: int, optional (default=None)
+        Random seed for reproducible subsampling.
```
jpaillard:
```suggestion
    random_state: int, optional (default=None)
        Random seed for reproducible subsampling. This will also 
        set the `random_state` parameter of the `vim` argument. 
```

jpaillard (approved):

bthirion (commented):
LGTM, but can you take care of the CI ? + latest comment by @jpaillard

GrituX:
Ok, I think I misunderstood his comment. I think the easiest is just to pass the input arguments directly from `cluster_imporance.fdr_selection` to `vim.fdr_selection`.

bthirion (approved):
LGTM, thx.

PR #716, opened 2026-07-15
[DOC] TabICL "advanced" example

### LINKED ISSUE #680: [DOC] Add example using TabICL

GrituX:
Restarting from #705 since PR included LOCI for some reason.
This closes #680.

### DIFF
--- a/docs/tools/references.bib
+++ b/docs/tools/references.bib
@@ -491,7 +491,7 @@ pages="39--68",
   year={2025}
 }
 
-@article{qu2026tabiclv2,
+@inproceedings{qu2026tabiclv2,
   title={{TabICLv2}: {A} better, faster, scalable, and open tabular foundation model},
   author={Qu, Jingang and Holzm{\"u}ller, David and Varoquaux, Ga{\"e}l and Le Morvan, Marine},
   booktitle={International Conference on Machine Learning},
--- a/docs/tools/references.bib
+++ b/docs/tools/references.bib
@@ -490,3 +490,10 @@ pages="39--68",
   booktitle={International Conference on Machine Learning},
   year={2025}
 }
+
+@article{qu2026tabiclv2,
+  title={{TabICLv2}: {A} better, faster, scalable, and open tabular foundation model},
+  author={Qu, Jingang and Holzm{\"u}ller, David and Varoquaux, Ga{\"e}l and Le Morvan, Marine},
+  booktitle={International Conference on Machine Learning},
+  year={2026}
+}
--- a/docs/tools/references.bib
+++ b/docs/tools/references.bib
@@ -483,3 +483,10 @@ pages="39--68",
   volume    = {76},
   year      = {2014}
 }
+
+@inproceedings{qu2025tabicl,
+  title={Tab{ICL}: {A} Tabular Foundation Model for In-Context Learning on Large Data},
+  author={Qu, Jingang and Holzm{\"u}ller, David and Varoquaux, Ga{\"e}l and Le Morvan, Marine},
+  booktitle={International Conference on Machine Learning},
+  year={2025}
+}
--- a/examples/plot_advanced_tabicl.py
+++ b/examples/plot_advanced_tabicl.py
@@ -24,9 +24,7 @@ from tabicl import TabICLRegressor
 
 dataset = fetch_california_housing()
 X, y, feat_names = dataset.data, dataset.target, dataset.feature_names
-X, y = resample(
-    X, y, replace=False, n_samples=1000, stratify=y, random_state=0
-)
+X, y = resample(X, y, replace=False, n_samples=500, stratify=y, random_state=0)
 X_train, X_test, y_train, y_test = train_test_split(
     X, y, test_size=0.2, random_state=0
 )
@@ -56,7 +54,6 @@ cfi = CFI(
     method="predict",
     n_permutations=5,
     loss=mean_squared_error,
-    n_jobs=-1,
 )
 cfi.fit(X_train, y_train)
 importance = cfi.importance(X_test, y_test)
--- a/examples/plot_advanced_tabicl.py
+++ b/examples/plot_advanced_tabicl.py
@@ -29,7 +29,7 @@ X_train, X_test, y_train, y_test = train_test_split(
     X, y, test_size=0.2, random_state=0
 )
 
-model = TabICLRegressor(device="cpu", kv_cache=True, n_jobs=-1)
+model = TabICLRegressor(device="cpu", kv_cache=True)
 model.fit(X_train, y_train)
 print(f"Model accuracy on the test data {model.score(X_test, y_test):.3}.")
 
@@ -54,6 +54,7 @@ cfi = CFI(
     method="predict",
     n_permutations=5,
     loss=mean_squared_error,
+    n_jobs=-1,
 )
 cfi.fit(X_train, y_train)
 importance = cfi.importance(X_test, y_test)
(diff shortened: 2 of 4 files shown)


examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,83 @@
...
+    {
+        "feature": feat_names,
+        "importance": importance,
+        "selected": selection,
+    }
+).sort_values("importance", ascending=False)
+
+import matplotlib.pyplot as plt
+
+loco.plot_importance()
```
bthirion: Can you plot the feature names to make the plot more interesting ?

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,83 @@
...
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.2, random_state=0
+)
+
+model = make_pipeline(
+    StandardScaler(), TabICLRegressor(device="cpu", kv_cache="repr")
+)
+model.fit(X_train, y_train)
+print(f"Model accuracy on the test data {model.score(X_test, y_test):.3}.")
+
```
bthirion, line 35: Is the accuracy better than that of standard models ?

bthirion (commented):
Looks great, thx ! Can we have error bare on the plot to make it more informative ?
Do we have any comment on the results with tabICL vs anything else ?

docs/tools/references.bib
```diff
@@ -483,3 +483,10 @@ @article{zhang2014confidence
   volume    = {76},
   year      = {2014}
 }
+
+@inproceedings{qu2025tabicl,
+  title={Tab{ICL}: {A} Tabular Foundation Model for In-Context Learning on Large Data},
+  author={Qu, Jingang and Holzm{\"u}ller, David and Varoquaux, Ga{\"e}l and Le Morvan, Marine},
+  booktitle={International Conference on Machine Learning},
+  year={2025}
+}
```
jpaillard, line 503: Also the second paper which introduces the regression model
```suggestion
}

@article{qu2026tabiclv2,
  title={{TabICLv2}: {A} better, faster, scalable, and open tabular foundation model},
  author={Qu, Jingang and Holzm{\"u}ller, David and Varoquaux, Ga{\"e}l and Le Morvan, Marine},
  booktitle={International Conference on Machine Learning},
  year={2026}
}
```

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,83 @@
...
+X, y = resample(
+    X, y, replace=False, n_samples=1000, stratify=y, random_state=0
+)
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.2, random_state=0
+)
+
+model = make_pipeline(
+    StandardScaler(), TabICLRegressor(device="cpu", kv_cache="repr")
+)
```
jpaillard: The model already includes basic preprocessing steps and averages predictions across slightly different preprocessing pipelines. 

```suggestion
model = TabICLRegressor(device="cpu", kv_cache="repr")
```

jpaillard (commented):
I think that having an example with PFI or CFI instead of LOCO and using `kv_cache=True` would illustrate the counter intuitive behaviour that appears with TabICL: `model.predict()` is more expensive than `model.fit()`, and this can be mitigated by kv-caching. 

Otherwise this looks good to me.

bthirion:
> I think that having an example with PFI or CFI instead of LOCO and using `kv_cache=True` would illustrate the counter intuitive behaviour that appears with TabICL: `model.predict()` is more expensive than `model.fit()`, and this can be mitigated by kv-caching.

Is this really the message we want to provide ? maybe yes, but this seems a bit advanced IMHO.
In any case, we should certainly use CFI and not PFI.

examples/plot_advanced_tabicl.py
GrituX: Should the example include a comparison to another model ? Idk how fair that is given that we had to greatly reduce the amount of samples due to GPU unavailability.

bthirion: Hm, you're right,  there is not enough interest in the comparison to complexify the example.

GrituX:
I agree that it's a bit advanced, but I don't think it's too bad to include it. The example is very basic otherwise and a little bit too straightforward...

bthirion (commented):
It takes 12 min to run ? Mamma mia, this will kill our CI. We need find a solution. E.g. run this example only when it changed or the underlying code changes, or only before releases.

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,82 @@
...
+dataset = fetch_california_housing()
+X, y, feat_names = dataset.data, dataset.target, dataset.feature_names
+X, y = resample(
+    X, y, replace=False, n_samples=1000, stratify=y, random_state=0
+)
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.2, random_state=0
+)
+
+model = TabICLRegressor(device="cpu", kv_cache="repr")
```
jpaillard:
```suggestion
model = TabICLRegressor(device="cpu", kv_cache=True)
```
IIUC, `kv_cache="repr"` uses less memory but provides less speedup than kv_cache=True (equivalent to `kv_cache='kv'`)

jpaillard (commented):
The time problem might be related to this warning: 
```
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
```
without a token, the download rate is crazy slow. Maybe we can add an HF token as a secret?

GrituX:
I added a CircleCI secret to hold the `HF_TOKEN`, but the example still takes 12 minutes to run.
I also looked at the previous doc build logs and couldn't find the warning that you mentionned @jpaillard . Where did you see it ?

GrituX:
The example now takes 56 seconds, which is much lower than originally, but still quite long given its simplicity. I think that the CI could be improved tbh. Currently, the doc is entirely built at each push on a PR that targets main by default. There is the possibility to build only the changes with the tag `[doc change]`. I think that this should be the default behavior, but this discussion is for another PR.

jpaillard:
Here you can see in the log the download from HF. 
image

bthirion (approved):
LGTM. 
yes, we'll have to revise our policy for CI handling.

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,86 @@
...
+        "importance": importance,
+        "selected": selection,
+    }
+).sort_values("importance", ascending=False)
+
+import matplotlib.pyplot as plt
+
+_, ax = plt.subplots()
+cfi.plot_importance(ax=ax)
+ax.set_yticklabels(df["feature"].values)
```
jpaillard, line 79: To make sure the the y labels are not cut. 
```suggestion
ax.set_yticklabels(df["feature"].values)
plt.tight_layout()
```

examples/plot_advanced_tabicl.py
```diff
@@ -0,0 +1,86 @@
...
+plt.show()
+
+# %%
+# As you can see from the timer under, even with 1000 samples and 9 features,
+# running TabICL on the CPU is slow. We recommend to use it on larger datasets,
+# be it in terms of samples and/or features and to run it on GPU for faster
+# inferences.
```
jpaillard: Actually, tabular foundation models are particularly good on small datasets compared to other models, thanks to their pre-training. So I would not say"We recommend to use it on larger datasets" 
```suggestion
# running TabICL on the CPU is slow. Using a GPU allows for an important speedup. 
```

jpaillard (approved):

examples/plot_advanced_tabicl.py
GrituX: I agree with you. My comment here is towards computation time, but my formulation is weird now that I read it. I'll remove it.

Issue #702, opened 2026-06-26
[ENH] Add M-plots


abaziree:
### Please confirm that you have searched existing issues in the repository.

- [x] Yes

### Please describe the feature that you would like to see improved or added.

Add implementation of M-plots.

### Describe the solution you would like to see implemented.

The idea is to average over the conditional distribution of the feature to get an idea of the evolution of the mean prediction depending on the values. 
It is a good compromise for handling categorical variables instead of ALE, even though the calculated importance may include indirect effects (e.g., when averaging conditionally on the number of rooms, the house size can still vary, which influences the resulting average value)

bthirion:
This makes sense to me BUT what do we say to the user ? We should make sure that we're not uselessly multiplying ways to plot variable influence, which makes the code harder to understand/use.
What I mean is that  maybe we should remove the PDP-only examples from the examples and e.g. replace it with M-plot.
And have a second example that compares ALE to M-plot and PDP (where PDP is only there for legacy).
Does that make sense ?

bthirion:
As a corollary, we need to have a narrative documentation section that plainly explain users what they should do in what situation.

abaziree:
> What I mean is that maybe we should remove the PDP-only examples from the examples and e.g. replace it with M-plot.

I think it could be better to join an M-plot example with the PDP examples rather than removing them because it's the same basically the same idea of averaging (marginal distribution for PDP and conditional distribution for M-plots). 

> And have a second example that compares ALE to M-plot and PDP (where PDP is only there for legacy)

Yep I agree

bthirion:
> I think it could be better to join an M-plot example with the PDP examples rather than removing them because it's the same basically the same idea of averaging (marginal distribution for PDP and conditional distribution for M-plots).

But people should always use M-plot by default ? If yes, it's a bad idea to keep PDP, because you put on  users the burden to decide what to use.

GrituX:
I think that if we add any of those methods to the codebase, we should add a base class `BaseVisualizer` as there is quite a lot of redundant code that could be regrouped.

PR #722, opened 2026-07-16
[FEAT] Visualization plot based on CPI

abaziree:
Visualization method :
1. A model $\mu$ predicts y based on X
2. We define $\eta_j$ which predicts $X_j$ based on $X_{-j}$
3. We plot $\mu(X) - \mu(X_{-j}, \eta_j(X_{-j}))$ averaged over each bin.


bthirion:
Thx ! Can you provide your preferred example with Z = X_1 ** 2 and X_2 correlated with X_1 ? That would help.

abaziree:
I added the example and here is the corresponding notebook to avoid compiling all the doc : [comparison_pdp_ale_cpi.ipynb](https://github.com/user-attachments/files/30118752/comparison_pdp_ale_cpi.ipynb)

abaziree:
The explanation of the test section in the code is that if we suppose $\mu(x) = f(x_j) + g(x_{-j})$ for simplification, then $\mu(x) - \mu(x_{-j}, \eta(x_{-j})) = f(x_j) - f(\eta(x_{-j})$. By assimilating a value of $x_j$ to its bin then we get a system to solve with unknowns $f_1, f_2, ..., f_K$ the values of f for each bin, and solving it with least squares give something which looks like ALE and without order constraint.

Also to answer what we talked yesterday, if $f(x) = \sum_{J \subset \{1,2,...,d \} } f_J(x_J)$ then we have 
$f_{J, ALE}(x_J) = f_J(x_J) + \sum_{u \subset J} h_u(x_u)$ for some functions $h_u$ which are of strictly lower order than $|J|$.

examples/plot_comparison_pdp_ale cpi.py
```diff
@@ -0,0 +1,129 @@
...
+_ = pdp.plot(X_test, features=features)
+
+# 2. ALE Plot
+ale = ALE(model, feature_names=X.columns)
+_ = ale.plot(X_test, features=features, grid_resolution=100)
+
+# 3. CPI Plot
+cpiplot = CPIPlot(model, feature_names=X.columns)
+_ = cpiplot.plot(X_test, features=features, grid_resolution=100)
+
```
bthirion, line 103: OK. In this example, the correlation between X2 and X1  is 0.985, which means that the model _cannot be properly identified. It work in you case, because you have picked a large sample size and added minimal noise on y (R^2 is unusually high). If you relax this a little bit, e.g. by using 
```
X2 = X1 + generator.normal(0, 0.5, n_samples)  # Strong correlation with X1
```
(the correlation remains 0.75) then the cpi model works properly.

bthirion (commented):
As I suspected your simulation setting is too extreme for conditional importance to work.

abaziree:
I added the examples with the california housing dataset. Is there a way to expose something such as correlation but between one variable and many others ?

bthirion:
You can create a synthetic variable that is the combination of others.

GrituX:
There is quite a lot of code here and in #702 that could be refactored with a base class BaseVisualization. But the real question before going there is to know whether we want to have all these different visualization methods or not.

PR #737, opened 2026-07-23
[MAINT] CI security fix

### LINKED ISSUE #735: [MAINT] CI security issues

- [x] Yes

The current CI relies on `request_pull_target` which has serious vulnerabilities since it runs in the context of the base repository and has access to Github secrets with write permissions. This is bad if a malicious PR ever comes and changes the workflow or executes malicious code with privileged access.

There are also other security issues where variables are accessed directly such as `github.event.pull_request.title` and used in `echo` commands which opens the door for injection attacks. Environment variables aren't used properly to avoid this too.

### Suggested fix:

The CI should avoid `request_pull_target` unless it really needs accessing Github secrets for PR automation. It is highly recommended to never check out and execute PR head code with elevated permissions. Builds and test should be executed with `pull_requests` to build and test the code and then a separate `workflow_run` can consume the artifacts (no code) to upload the documentation, and test reports.

For the injection attack issue, the fix is to use environment variable in quotes:
```yaml
- name: Print PR title
  env:
    PR_TITLE: ${{ github.event.pull_request.title }}
  run: echo "$PR_TITLE"  # safe way
  run: echo $PR_TITLE  # unsafe
```

_No response_

GrituX:
Untrusted PR code is now run under more restrained security with a switch from `pull_request_target` to `pull_request`, ensuring that secrets cannot be read by malicious code and blocks any workflow change. This change imposes that both files `ci.yml` and `ci_maintenance.yml` be split in two:
- Everything before the publish call is done in a `pull_request` workflow and remains mainly unchanged, which executes untrusted build and tests.
- The publishing part requires access to secrets and is done in a separate workflow now `ci_publish_call.yml` that gets the secrets on its own, and allows to get rid of `secrets: inherit` that opens security risks

There are other security fixes in other files to avoid code injection from different sources.

`.circleci/config.yml` has been changed due to a weird behavior: previously the ssh key was systematically loaded into the ssh agent before checking whether to deploy the documentation or not (doc deployed only on release branches). I moved the check before adding the ssh-key.


GrituX:
I think that the added intermediate file `ci_publish_call.yml` and `call_publish_result.yml` can be condensed into a single file. I'm working on it.

Also CircleCI doesn't seem happy about the change, I'm figuring this out too.

GrituX:
It seems that it works. The only thing now that we can't see on this branch: the publishing part of the CI will not show up until it's on the main branch. I had to rework the way this workflow is called and had to give a name to it. Since workflows are called from registered workflows on the main branch, and the new one isn't registered yet, well it's never called then. And so the only way to have that, is to merge the code.

bthirion:
Thx. Can you clarify what "the publishing part of the CI" relates to ? It's unclear to me.

GrituX:
It's the `ci_publish_call.yml`file. Basically it's all the CI part that gets all the test results, test coverage, reports, build outputs, and uploads them.

bthirion (approved):
Hard to me to tell whether everything is technically right, but it seems to address the intended goal.

Remi-Gau (approved):
that globally makes sense to me: I have not checked all the nitty gritty though

GrituX:
I will merge this tomorrow and hopefully all goes well.

.github/workflows/ci_maintenance.yml
Remi-Gau:
```suggestion
          if [ "${{ github.event_name }}" == "pull_request" ]; then
```

sorry just did a search for pull_request_target

shouldn't this be changed?

.github/workflows/ci.yml
Remi-Gau:
```suggestion
          if [ "${{ github.event_name }}" == "pull_request" ]; then
```

Remi-Gau: there is also a similar thing in ci_test.yml

.github/workflows/ci_maintenance.yml
GrituX: Yes, totally. Thx for the check

.github/workflows/ci.yml
GrituX: The change is also necessary for `ci.yml` as well as `ci_test.yml`, thanks !

GrituX:
Ok, now the CI seems to be working as intended. The last thing that does not get called is the workflow `ci_publish_call.yml`, but that's normal since it requires to be merged into the main branch.

GrituX:
The CI is working now, I tested it with PR #743 .
However, the publishing of the test results, doc generation, and code coverage computation are not triggered after a push on main. Should we change this ? The former CI used to do this.

GrituX:
I'm opening an issue and PR #744 , and we can decide later whether we merge this or not. IIUC, the publishing part is necessary so that the documentation on the website is updated. So the fix should be pretty useful.

Remi-Gau:
doc generation of the main branch is very useful to have at any times
code coverage: is it not uploaded to codecov
test results: depends if you have ever needed to actually check the artefacts

GrituX:
That's what I thought. PR #744 should fix all this.

PR #425, opened 2025-09-15
Very high dim

bthirion:
One setp toward #306 : populating the high dimension documentation module.

PR #739, opened 2026-07-28
[ENH] Speed up tests

### LINKED ISSUE #700: [ENH] Speed-up tests
Some tests are run on ~1000 samples and 200 features when we don't actually need that much. It could be good to have a global check on tests and try to improve their runtime.

GrituX:
I managed to reduce by nearly 40% the runtime of tests on my machine by reducing the amount of generated samples and/or features, and other small fixes. This should help the CI a bit.

On the side, I have a few remarks:
- CFI/test_group fails because there is no UserWarning, which is very weird. I can't understand why it doesn't emit the warning.
- Non-regression tests aren't supposed to stay forever generally. Currently there are these:
    - DCRT/test_d0crt_no_regression_variance_fix from PR #649 . This one is still pretty new, but it has not failed once since the change. This one is a pure regression test, and could be removed at some point.
    - EnCluDL/test_cludl_independence (which I also wrote for to EnsembleImportance and ClusterImportance) from PR #425 and recently reworked in PR #707. I think this should be kept in general as it tests CluDL's selection on repeated calls.


GrituX:
I'm looking into failed tests

GrituX:
I relaxed the constraint on the EnCluDL/test_cludl_independence test as it failed quite often recently and is supposed to be removed at some point.

jpaillard:
One thing to keep in mind, which caused us some trouble in the past, is that the methods tested are stochastic. 
So the tests might pass most of the time but still fail once in a while due to the method's "intrinsic randomness": for instance, when a specific sample draw leads to more false discoveries. That's quite annoying because in PRs it becomes difficult to figure out if the failure comes from the modifications or just happens by chance. 

Keeping n "large enough" has been a good way to make the tests more robust. We don't have a clear criterion for robustness (maybe that would be a good thing for the dev. doc). I personally use the rule of thumb that it should pass more than 99% of the time (repeated with different seeds). 

Also, regarding the dimension, for methods that control the **rate** of false discoveries (like knockoffs), increasing the dimension usually improves robustness but also requires increasing n.

Test robustness is also the reason why, for many tests, we actually repeat them 10 times and check if the average FDP, power ... over repetitions corresponds to what we expect.

GrituX:
I tried to keep a good balance between what stochasticity imposes and performance. I ran all tests 20 times on my machine to check which one failed. But apparently it didn't catch everything before I pushed the code. I can run more loops to evaluate this.

GrituX:
I ran all tests 500 times, and the only one that failed once in a while is the EnCluDL/test_cludl_independence test. So I think it should be fine.

bthirion:
> I ran all tests 500 times, and the only one that failed once in a while is the EnCluDL/test_cludl_independence test. So I think it should be fine.

We probably need to mend it so that failures become rarer.

GrituX:
I'm looking for a solution, I'll open an issue and PR for that.
All other tests are fine, so the changes are open for review.

test/conftest.py
```diff
@@ -95,8 +95,8 @@ def data_generator(
         shuffle=False,
         seed=seed,
     )
-    important_features = np.where(beta != 0)[0]
-    not_important_features = np.where(beta == 0)[0]
+    important_features = np.where(beta)[0]
+    not_important_features = np.where(~beta)[0]
```
bthirion: In most cases, these `np.where()` are useless, as we can simply work with binary vectors.

test/test_conditional_feature_importance.py
```diff
@@ -632,12 +639,15 @@ def test_groups_warning(self, data_generator):
             method="predict",
             features_groups=subgroups,
             feature_types="auto",
+            n_jobs=-1,
         )
         cfi.fit(X, y)
+        cfi.importance(X, y)
+        print("lol")
```
bthirion: ?

test/test_desparsified_lasso.py
```diff
@@ -280,7 +283,7 @@ def test_exception():
         tol=1e-4,
         max_iter=5000,
         random_state=1,
-        n_jobs=1,
+        n_jobs=-1,
```
bthirion: Isn't it a bit risky ?

test/test_desparsified_lasso.py
```diff
@@ -493,7 +496,7 @@ def test_group_reid_2():
     )  # covariance matrix of time
     cov = 1.0 * corr
 
-    lasso_cv = MultiTaskLassoCV(n_jobs=1).fit(X, y)
+    lasso_cv = MultiTaskLassoCV(n_jobs=-1).fit(X, y)
```
bthirion: Same for those. I think you want to avoid `n_jobs=-1` in tests

bthirion (commented):
Great ! Thx

test/test_conditional_feature_importance.py
GrituX: oops forgot to delete that

test/test_desparsified_lasso.py
GrituX: The parallelization is done on an external library that is also tested, so this should be robust enough.

test/conftest.py
GrituX: I agree. however they're useful here as most test methods need features indexes from this fixture

test/test_desparsified_lasso.py
GrituX: What's the problem with that ?

bthirion: Well, if you run tests on your machine in bg, you suddenly have all your cpus used by the tests...

GrituX: I agree, but I think it's also important that tests run as fast as possible online.

Remi-Gau: FYI you are using pytest xdist to parallelize test and from what I can see you are letting xdist decide on the optimal number of CPU to use in CI with the option `-n auto` 

In practice I think it means you are using 2 workers to run tests in parallel
https://github.com/mind-inria/hidimstat/actions/runs/30903322176/job/91973356856?pr=739#step:6:27

So I think that you do not have any CPU available to parallelize with n_jobs in CI.

When running things locally I tend to go with `-n` between 10 or 20 depending on how many tests I want to run in parallel.

When you need to actually test the parallelization of some code with n_jobs in nilearn we have flagged those tests with @pytest.mark.single_process marker (also used for tests that may become flaky with pytest xdist): then we run these marked tests in a separate job in CI with -n 1 to avoid xdist parallelization

that was my long 2 cents on this

Remi-Gau:
also... having an entire test suite that runs in under 3 minutes make me really envious!!! 😜

Remi-Gau:
> t still fail once in a while due to the method's "intrinsic randomness": for instance, when a specific sample draw leads to more false discoveries.

not looked into this but can this not be pinned with a proper RNG?

test/test_desparsified_lasso.py
GrituX: I was about to suggest to switch from `n_jobs=-1` and use xdist to parallelize everything, but it seems that I overlooked that it was already included online for the CI.
I'll revert this then, thanks !

test/conftest.py
```diff
@@ -95,9 +95,9 @@ def data_generator(
         shuffle=False,
         seed=seed,
     )
-    important_features = np.where(beta != 0)[0]
-    not_important_features = np.where(beta == 0)[0]
-    return X, y, important_features, not_important_features
+    important_features_ids = np.arange(X.shape[1])[beta]
```
bthirion: why not simply `important features = beta != 0`

bthirion (commented):
Almost there, thx.

jpaillard:
> > t still fail once in a while due to the method's "intrinsic randomness": for instance, when a specific sample draw leads to more false discoveries.
> 
> not looked into this but can this not be pinned with a proper RNG?

If we do that, the test will depend on the pinned random seed, and any change in the random state (like simply changing the order of random function calls) might break the test without changing the implementation logic.  I think that we don't want that, the test should detect changes in the behavior of the method "on average," not for a specific random seed.

test/conftest.py
GrituX: Quite a few tests actually need the feature ids and not just a mask

bthirion: Well, but unless I'm wrong it's historical: it's an old pattern that is somewhat pervasive across th codebase/tests, but we should get rid from it wherever it's possible.

GrituX:
> Well, but unless I'm wrong it's historical: it's an old pattern that is somewhat pervasive across th codebase/tests, but we should get rid from it wherever it's possible.

I made the change to reduce unnecessary compute for tests that didn't use feature ids

bthirion (commented):
one comment pending, LGTM otherwise !

test/conftest.py
GrituX: I switched from this pattern to boolean masking !

test/test_conditional_feature_importance.py
```diff
@@ -676,17 +681,19 @@ def test_groups_warning(self, data_generator):
             feature_types="auto",
         )
         cfi.fit(X, y)
+        cfi.importance(X, y)
+        print("lol")
```
bthirion:
```suggestion
```

GrituX:
I think I went through all comments, but maybe I forgot one ? Which one is it ?

bthirion (approved):
LGTM, thx.

PR #740, opened 2026-07-30
[ENH] Unified handling of groups with pandas DataFrames

### LINKED ISSUE #721: [BUG] Handling of pandas dataframes and numpy arrays

- [x] Yes

### Describe your issue.

Currently the behaviour is inconsistent between methods: 
 - CFI converts DataFrames to array before passing them to estimators
 - LOCO/LOCI keep the dataframes which leads to an error since the groups are internally defined as array indices, not columns `KeyError: "None of [Index([0], dtype='int64')] are in the [columns]"`

I would suggest to rework the handling of groups so that we keep `pd.DataFrames` (otherwise estimator pipelines fitted on dataframes will be unhappy) and change the `self.features_groups_` values from array indices to dataframe columns whenever the input is a dataframe. 

### Reproducing Code Example

```python
from sklearn.linear_model import LinearRegression
import pandas as pd
from hidimstat import LOCI

df = pd.DataFrame({"X_1": [1, 2, 3], "X_2": [4, 5, 6]})
y = [0, 1, 0]
estimator = LinearRegression()
estimator.fit(df, y)
loci = LOCI(estimator=estimator)
loci.fit(df, y)
```

### Error message

```shell
KeyError: "None of [Index([0], dtype='int64')] are in the [columns]"
```

```shell
0.3.2
```

GrituX:
When a pandas DataFrame is given as input, now the vim handles the column names of the DataFrame without any conversion to numpy. `self.features_groups_` now directly contains pandas columns names. Grouping can directly be done on column names.


GrituX:
I opened an issue #741 for the dataset download fail. I'll launch another build on the CI later

src/hidimstat/_utils/utils.py
```diff
@@ -76,6 +77,17 @@ def _generate_group_mask(array_size, indexes, selected=True):
     return mask
 
 
+def _get_array_cols(X, columns_ids, drop=False):
+    """
+    Retrieves columns from X depending on type numpy array or pandas DataFrame, or drops them from X.
```
bthirion: can we limit lines to 80 characters ? (for readability)

bthirion (commented):
LGTM overall.

src/hidimstat/_utils/utils.py
GrituX: yes ofc

src/hidimstat/samplers/conditional_sampling.py
jpaillard, line 113:
```suggestion
               np.asarray(y).ravel()
```
The feeding a dataframe as X with `features_groups=None` would break without that. 
It was also not caught by any test; maybe we want to have a test for that.

src/hidimstat/shapley_additive_global_importance.py
```diff
@@ -49,7 +50,10 @@ def _sage_value_function(
     )  # (n_permutations, n_samples, n_features)
     for perm_idx in range(n_permutations):
         for col in complement:
-            X_sampled[perm_idx, :, col] = rng.permutation(X[:, col])
+            X_col = _get_array_cols(X, col)
+            X_sampled[perm_idx, :, col] = rng.permutation(X_col).reshape(
+                -1,
+            )
 
     X_sampled_batch = X_sampled.reshape(-1, n_features)
```
jpaillard, line 58:
```suggestion
    X_sampled_batch = X_sampled.reshape(-1, n_features)
    if isinstance(X, pd.DataFrame):
        X_sampled_batch = pd.DataFrame(X_sampled_batch, columns=X.columns)
```
same as CFI, use columns names to remove a warning

jpaillard (changes requested):

src/hidimstat/samplers/conditional_sampling.py
bthirion: Can you add a  test ?

test/test_conditional_feature_importance.py
```diff
@@ -237,17 +237,47 @@ def test_group(data_generator):
...
+        n_permutations=20,
+        method="predict",
+        features_groups=None,
+        feature_types="auto",
+        random_state=0,
+        n_jobs=1,
+    )
+
+    # Check that the conditional samplers properly detect this is not a multi-output setting.
+    cfi.fit(X_df)
```
jpaillard:
```suggestion
    cfi.fit_importanec(X_df, y)
```

To make sure it's robust, let's also test the sampling part (not only the fitting)

test/test_conditional_feature_importance.py
```diff
@@ -237,17 +237,47 @@ def test_group(data_generator):
...
+    assert all(not model.multioutput_ for model in cfi._list_imputation_models)
+
+    # Check same thing when X is a np.ndarray
+    cfi.fit(X)
```
jpaillard:
```suggestion
    cfi.fit_importance(X, y)
```

jpaillard (approved):
Almost good, thx. 
I only suggest to also test the importance call for both arrays and dataframes.

GrituX:
I added a simple test for the sampling part as the comparison between important and unimportant feature is done several times in other tests.

PR #751, opened 2026-08-12
[BUG] allow unfitted estimators for CFI

### LINKED ISSUE #750: [BUG] CFI does not support un-fitted estimators

- [x] Yes

### Describe your issue.

The `BasePerturbation` class currently supports unfitted estimators by running the `_initial_fit` method https://github.com/mind-inria/hidimstat/blob/59737f51e5261581a6f8411778d164d46f66abbe/src/hidimstat/base_perturbation.py#L126

However, the `y` argument is deleted in CFI.fit() which prevents the `BasePerturbation._initial_fit()` from being properly used. https://github.com/mind-inria/hidimstat/blob/59737f51e5261581a6f8411778d164d46f66abbe/src/hidimstat/conditional_feature_importance.py#L120-L121

### Reproducing Code Example

```python
model = LinearRegression()

cfi = CFI(
   estimator=model, 
)
cfi.fit(X_train, y_train)
```

### Error message

```shell
ValueError: This LinearRegression estimator requires y to be passed, but the target y is None.
```

```shell
latest
```

jpaillard:


test/test_conditional_feature_importance.py
```diff
@@ -943,3 +943,19 @@ def test_cfi_cv(data_generator):
...
+)
+def test_unfitted_estimator(data_generator):
+    """Test CFI with an unfitted estimator"""
+    X, y, _, _ = data_generator
+
+    regression_model = LinearRegression()
+
+    cfi = CFI(estimator=regression_model)
+    fitted_cfi = cfi.fit(X, y)
+    assert isinstance(fitted_cfi, CFI)
```
GrituX: Since the goal here is to test that _initial_fit is properly called on `regression_model` wouldn't it be better to have something like `assert check_is_fitted(cfi.estimator_)` ?

jpaillard: yep, that makes more sense

GrituX:
All good for me, thanks for that !

GrituX (approved):

bthirion (approved):
LGTM, thx.

PR #759, opened 2026-08-18
[TEST] Fixed tests with deterministic seed

### LINKED ISSUE #758: [TEST] Deterministic seed for tests
A few tests run with a deterministic seed such as follows:

https://github.com/mind-inria/hidimstat/blob/138f48b8288ebc1d982f72f06ca60d6e6814e301/test/test_desparsified_lasso.py#L74-L99

This should be replaced with the available fixture:

https://github.com/mind-inria/hidimstat/blob/138f48b8288ebc1d982f72f06ca60d6e6814e301/test/conftest.py#L25-L32

This pattern can be found in the following test files:

- test_desparsified_lasso.py
- test_ensemble_clustered_importance.py
- test_ensemble_importance.py
- test_knockoff.py

GrituX:
I suggested a change from deterministic seeds to using the rng fixture and then `rng.integers(low=0, high=500, size=10)` for seed loops.
However, this currently breaks one test `test_enclud_spatial` in `test_ensemble_clustered_inference.py` for some reason. I'm trying to figure out what's the problem.

### DIFF
--- a/test/test_ensemble_clustered_inference.py
+++ b/test/test_ensemble_clustered_inference.py
@@ -195,10 +195,10 @@ def test_encludl_spatial(rng):
             selected=selected,
             ground_truth=beta,
             roi_size=roi_size,
-            spatial_tolerance=2,
+            spatial_tolerance=3,
             shape=shape,
         )
-        fp_list.append(fdp)
+        fp_list.append(int(fdp > 0))
         power_list.append(power)
     assert np.mean(power_list) >= 0.5
     assert np.mean(fp_list) <= fwer + tol
--- a/test/test_ensemble_clustered_inference.py
+++ b/test/test_ensemble_clustered_inference.py
@@ -46,7 +46,7 @@ def spatially_relaxed_fdp_power(
     return fdp, power
 
 
-def test_cludl_spatial():
+def test_cludl_spatial(rng):
     """
     Test CluDL on a 2D spatial simulation. Testing for support recovery methods using
     clustering is challenging as clusters that intersect the true support can also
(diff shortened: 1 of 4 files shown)


bthirion:
Could be simple randomness on these rather complex estimators...

GrituX:
Indeed, but with this change the FDP jumps to 0.7 which is a fairly dramatic rise.

bthirion:
> Indeed, but with this change the FDP jumps to 0.7 which is a fairly dramatic rise.

oops, indeed

GrituX:
I think the test was wrong, in that it checked a repetition of the algorithm making at least one false discovery. I changed it to an assertion on the fdp being under a certain threshold.

test/test_ensemble_clustered_inference.py
```diff
@@ -198,13 +198,13 @@ def test_encludl_spatial():
             spatial_tolerance=2,
             shape=shape,
         )
-        fp_list.append(int(fdp > 0))
+        fp_list.append(fdp)
```
bthirion: What is controled here is the fwer, which is  basically the rate of image-level tests with **at least one** false positive. This is why the previous test was actually correct. 

What you can do instead is increase spatial_tolerance.

GrituX: Indeed, I went through this a bit too quickly.
I'll do that instead.

bthirion (approved):
LGTM, thx.

jpaillard (approved):
Thanks I think it improves the tests. 

However, it's unsatisfying that we need to change the tests (increase n here), because they are brittle. for the CI to pass. Here, it is not a problem since only the seed is changed, but it will be problematic when this occurs in a PR that modifies a method.

PR #761, opened 2026-08-19
[TEST] Improved fixture usage, reduce test code repetition

### LINKED ISSUE #760: [TEST] Test fixture not used to full potential
There is a lot of data generation code that could be replaced by the `data_generator` fixture, which would reduce maintenance cost, and potentially speed up tests too. Here are the following tests that are concerned:
- desparsified lasso
- dcrt
- encluinf
- ensemble importance
- knockoff
- loci / loco
- PFI

GrituX:
I went through the entire test codebase to find spots where code was heavily duplicated and reduce it with existing fixtures, new fixtures or helper functions.


test/test_base_variable_importance.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pytest.fixture
-def set_100_variable_sorted(rng):
+def set_100_variable_sorted(rng, create_BVI):
```
bthirion:
```suggestion
def set_100_variable_sorted(rng, create_bvi):
``` I think we should document this argument

File: test/test_base_variable_importance.py [hunk 2]
```diff
@@ -30,6 +30,11 @@ def set_100_variable_sorted(rng):
     return vi
 
 
+@pytest.fixture(scope="function")
+def create_BVI():
```
bthirion:
```suggestion
def create_bvi():
```

test/test_conditional_feature_importance.py
```diff
@@ -117,26 +128,27 @@ def cfi_test_data():
...
 ##############################################################################
 ## tests cfi on different type of data
+PARAM_NAMES = "n_samples, n_features, n_targets, support_size, rho, seed, value, signal_noise_ratio, rho_serial"
 PARAM_EXACT = [
-    (150, 200, 10, 0.0, 42, 1.0, np.inf, 0.0),
-    (150, 200, 10, 0.0, 42, 1.0, 10.0, 0.0),
-    (150, 200, 10, 0.0, 42, 1.0, 10.0, 0.2),
+    (150, 200, None, 10, 0.0, 42, 1.0, np.inf, 0.0),
+    (150, 200, None, 10, 0.0, 42, 1.0, 10.0, 0.0),
+    (150, 200, None, 10, 0.0, 42, 1.0, 10.0, 0.2),
```
bthirion: Why this change ?

bthirion (commented):
LGTM overall, thx !

test/test_conditional_feature_importance.py
GrituX: I added an extra argument `n_targets` to the `data_generator` fixture so that it can be used in extra locations.

bthirion (approved):
LGTM, thx.

test/test_base_variable_importance.py
```diff
@@ -30,6 +37,11 @@ def set_100_variable_sorted(rng):
     return vi
 
 
+@pytest.fixture(scope="function")
+def create_bvi():
+    return BaseVariableImportance()
```
jpaillard, line 42: Is this really necessary, given that it's a single line of code and makes the test less transparent?

test/test_distilled_conditional_randomization_test.py
```diff
@@ -112,6 +112,20 @@ def generate_binary_classif_dataset(n=100, p=10, seed=0):
     return X, y, beta
 
 
+@pytest.fixture
+def generate_regression_data(noise):
+    X, y = make_regression(
+        n_samples=100, n_features=10, noise=noise, random_state=42
+    )
+    return X, y
```
jpaillard, line 120: Can't we use the already existing `data_generator` fixture?

jpaillard (approved):
Thx I think this simplifies a lot the tests.

test/test_distilled_conditional_randomization_test.py
GrituX: Yes, it avoids code duplication.

test/test_base_variable_importance.py
GrituX: It's more of a convention, it doesn't change much in execution time. For tests, this actually slightly improves readbility imo: the fixture is given as parameter of the test, so this is an indication of what the test requires. But if you feel like it makes the code less transparent, I can remove that.

bthirion:
We can merge whenever the doc builds.

GrituX:
Yup, i'm waiting for #767 to be resolved to merge this.

bthirion (commented):
LGTM, but please have an eye on the cost of running tests.

PR #774, opened 2026-09-10
[BUG] LOCI now uses marginal probability in binary classif

### LINKED ISSUE #772: [BUG] LOCI could use class probabilities for binary classification

- [x] Yes

### Describe your issue.

In LOCI, for the binary classification case, we currently use a "chance" model that predicts the majority class. I think this is not the right baseline as it performs much worth that a model predicting the probability of y=1 when using the log-loss. 
As a result, it gives overly optimistic importance scores. I suggest switching to the probability (which we already use for the multiclass case). 

### Reproducing Code Example

https://github.com/mind-inria/hidimstat/blob/ed340a23266f75d43c213928311d13b4cd97d803/src/hidimstat/leave_one_covariate_in.py#L159-L161

### Error message

```shell
Silent, does not crash
```

```shell
latest
```

GrituX:
I switched to marginal probabilty of the positive class as the baseline for LOCI in binary classification:
```python
if self.method in ["predict_proba", "decision_function"]:
  values, counts = np.unique(y, return_counts=True)
  # We take the marginal probability in any case.
  # Binary classification, shape of y is (n_samples,)
  if len(values) == 2:
      self._baseline_mean = counts[1] / y.shape[0]
  # For multiclass classification, shape of y is (n_samples, n_classes)
  else:
      self._baseline_mean = counts / y.shape[0]
```

I also removed the redundant line.

### DIFF
--- a/src/hidimstat/leave_one_covariate_in.py
+++ b/src/hidimstat/leave_one_covariate_in.py
@@ -103,11 +103,11 @@ class LOCI(BasePerturbation):
         )
         if self.method in ["predict_proba", "decision_function"]:
             values, counts = np.unique(y, return_counts=True)
-            # Binary classification
+            # We take the marginal probability in any case.
+            # Binary classification, shape of y is (n_samples,)
             if len(values) == 2:
-                values, counts = np.unique(y, return_counts=True)
-                self._baseline_mean = values[np.argmax(counts)]
-            # For multiclass classification, we take the marginal probability.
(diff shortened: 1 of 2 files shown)


jpaillard (approved):
thx

bthirion (commented):
If you change, the behavior, you need to add a test.

GrituX:
Done

bthirion (approved):
LGTM, thx.

Issue #768, opened 2026-09-08
Hidimstat Sprint Day 09/26


GrituX:
We are holding a Sprint Day at Inria next Monday to talk about the next steps for version 0.5.0 ! I'm opening this issue to throw a few ideas, discuss and prepare topics for the day. Here are a few things that may be of interest:
- Graphs: since Aurélien finished his internship, we may want to take a bit of time to discuss ALE, M-plots, etc to know whether we want to integrate new visualization methods, or refine existing ones.
- Documentation: it could be a good opportunity to discuss in person the example gallery rework that has been left in standby. Maybe there are other documentation needs that I'm missing too ?
- New methods: Do we want to include other methods ? Are there any gaps that need to be filled ?
- Auto-report: Maybe we could think about having an auto-generated report that plots variable importance, suggests variable selection given a certain input ?

Current attendees are now: @bthirion @jpaillard @AngelReyero @GrituX 

Feel free to suggest other ideas, discuss current ones, or invite other people

bthirion:
@jeromedockes @rcap107

bthirion:
Some comments on the user guide:
* Some snippets in the user guide are not runnable, e.g. because X, and y are not defined. Having a one-liner to add them (+ imports) would not be a big deal and would make the snippets more valuable imho.
* 1.1.2. Could be **Controled variable selection
* More generally, I see lots of improvements to perform on the user guide

jeromedockes:
@glemaitre

GrituX:
The day is going to be held at the Turing building of INRIA, in the David Hilbert room starting from 9:30 am !

GrituX:
User guide: 
- [x] Check that code snippets in User guide run (autonomous)
- [x] Fill section 1.2 stat guarantees
- [x] Homogenize "style" of section organization
- [x] Fill section 3.4 LOCO
- [x] Fill section 4.1 LOCI
- [ ] Fill section 5 PDP
- [ ] Fill section 5 ALE
- [ ] Fill section 6 on groups
- [ ] add an "overview" section for a complete pipeline structure

bthirion:
I can start a PR on 1.2 stat guarantees

PR #793, opened 2026-09-14
[DOC] User guide style homogenization

GrituX:
I homogenized all user guide files. However, either we have section pages that contain all the documentation, either we have the section navigation on the left that displays all sections and subsections, we can't have both ... So a choice is to be made here: visibility against navigation. I think visibility is primordial in documentation. Let me know what you think !
Issued from #768


docs/src/slm_methods/model_x_knockoffs.rst
```diff
@@ -0,0 +1,129 @@
...
+
+The Model-X Knockoffs method :footcite:t:`candes2018panning` is a method for variable
+selection that controls the false discovery rate (FDR). It is particularly useful in
+high-dimensional settings since its default implementation leverages sparse regression
+with the Lasso.
+
+
+.. figure:: ../generated/gallery/examples/images/sphx_glr_plot_knockoffs_wisconsin_001.png
+    :target: ../generated/gallery/examples/plot_knockoffs_wisconsin.html
+    :align: center
```
bthirion: Does not render. Besides, I think that this kind of figure is useless.

bthirion (commented):
Thx !
General point: currently we have references in the middle of the pages, which is not great. Would it be possible to have them only at the end of each section ?

docs/src/slm_methods/model_x_knockoffs.rst
GrituX: Let me know if you think other figures are useless.
But yes, there are 6 figures that are not being displayed even though the path is correct. Something's off

GrituX:
I decided to get rid of the failing figures as they weren't really explained or used in the text.

GrituX:
This should also resolve the issues on the main branch

docs/src/concepts/general_concepts.rst
```diff
@@ -120,12 +118,7 @@ features without such control is not valid. Different forms of guarantees can
 be employed, such as controlling the type-I error or the False Discovery Rate.
 This step is directly related to the task of Variable Selection.
 
-
-
-
-
-
 References
-----------
```
bthirion, line 129: It's a bit weir to have 1.2, 1.4 and 1.6 as reference sections. Is it possible to have an unnumbered section ?

bthirion: Or simply have references as section 1.1.7, 1.2.4 etc. ?

bthirion (commented):
LGTM besides reference section numbering

docs/src/concepts/general_concepts.rst
GrituX: I moved the `References` sections down a section level, i.e. as 1.1.7 etc.

bthirion (approved):
LGTM, thx !

docs/src/slm_methods/model_x_knockoffs.rst
Remi-Gau: that should be fixed in the PR that transfer the doc to github: if you need to refer to figures of the examples somewhere else in the doc, you need to add a comment in the example (I fixed it in the doc transfer PR) to say how many dummy figures should be generated when the example is not run so that the reference has an actual image file to point to.

hope that helps

jpaillard (approved):

docs/src/slm_methods/model_x_knockoffs.rst
bthirion: To me a figure is useless when it is just a vague reference: if it does not point to explicit code or example, it's just illustrative.

PR #798, opened 2026-09-15
[MAINT] transfer doc build to github

### LINKED ISSUE #786: [MAINT] doc not built in fork PR

- [x] Yes

See #781, but documentation build is not being launched on CircleCI after tests are completed on Github Actions.

### Suggested fix:

_No response_

_No response_

Remi-Gau:
- closes #786 
 - uses the nilearn workflow to build the doc: https://nilearn.github.io/dev/ci.html#documentation-build-and-deploy


examples/plot_cfi.py
```diff
@@ -110,3 +110,6 @@
 # References
 # ----------
 # .. footbibliography::
+
+
+# sphinx_gallery_dummy_images=2
```
Remi-Gau, line 115: needed so sphinx-gzllery generates dummy images when the example is not run: would otherwise tigger a warning that a target image is missing and this would fail the build

.circleci/config.yml
```diff
@@ -1,146 +1,42 @@
...
-      - persist_to_workspace:
-          root: docs/_build/html
-          paths: .
-      - save_cache:
-          # cache some library
-          key: saved-cache
-          paths:
-            - /home/circleci/nilearn_data
-            - /home/circleci/sklearn_data
-            - /home/circleci/mne_data
```
Remi-Gau, line 75: the old ci was caching data so let's add this to github too

.circleci/config.yml
```diff
@@ -1,146 +1,42 @@
...
+                path: doc/_build/html/
+                destination: dev
+        # Persists the generated documentation, so that it
+        # can be attached and deployed in the "deploy" job
+        -   persist_to_workspace:
+                root: doc/_build
+                paths:
+                -   html
 
-  get-tests-reports:
```
Remi-Gau, line 77: how frequently are we checking the tesrt reports on circle CI?

.github/workflows/ci_publish_call.yml
```diff
@@ -55,92 +55,6 @@ jobs:
           echo "BRANCH_NAME=main" >> $GITHUB_OUTPUT
           echo "SHOULD_PUBLISH=true" >> $GITHUB_OUTPUT
 
-  merge_and_upload_test_data:
```
Remi-Gau, line 58: assuming this is not needed: to revert if any of the maintainers frequently go to circle ci to view the tests reports

docs/Makefile
```diff
@@ -2,9 +2,7 @@
 #
 
 # You can set these variables from the command line.
-N_JOB         = auto
-SPHINXOPTS    = -j ${N_JOB} --fail-on-warning --keep-going --nitpicky --verbose
-# -j/--jobs:		    N processes in parallel (auto=number of CPUs)
```
Remi-Gau, line 7: the html theme does not support parallel build

.github/workflows/ci_publish_call.yml
jpaillard: I have not seen the test reports in PRs for some time. Would it make sense to have the test report from codevov in the PR?

jpaillard (commented):
Thanks! That will help review the doc PRs

.circleci/config.yml
GrituX: Rarely. But I think it's still nice to keep, in case we want to check performances.

.github/workflows/ci_publish_call.yml
GrituX: Now that you mention it, it's been a while since we've had a report from Codecov

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,296 @@
...
+        #     │ ┌───────────── hour (0 - 23)
+        #     │ │ ┌───────────── day of the month (1 - 31)
+        #     │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+        #     │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+        #     │ │ │ │ │
+        #     │ │ │ │ │
+        #     │ │ │ │ │
+        #     * * * * *
+        # Runs at 00:00, on day 1 of every 3rd month
+    -   cron: 0 0 1 */3 *
```
GrituX: I don't think that we currently need a regularly scheduled doc building. The doc is built for every push on main, imo that is more than enough right now

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,296 @@
...
+        -   name: Install apt packages
+            run: |
+                sudo -E apt-get -yq update
+                sudo -E apt-get -yq --no-install-suggests --no-install-recommends install \
+                    dvipng texlive-latex-base texlive-latex-extra
+
+            # Set up and launch a virtual browser needed for one example to run
+            # without stalling the job. The example launches an html in the browser.
+        -   name: Set up display server for virtual browser
+            run: Xvfb -ac :99 -screen 0 1280x1024x16 > /dev/null 2>&1 &
```
GrituX: I don't understand, why do we need that ?

tools/documentation/github/build_type.py
```diff
@@ -0,0 +1,247 @@
...
+    for filename in changed_files:
+        if (
+            "examples/" in filename
+            and "plot_" in filename
+            and filename.endswith(".py")
+        ):
+            print(f"Checking example {filename} ...")
+            pattern_parts.append(Path(filename).name)
+
+    build_type = "html-noplot"
```
GrituX, line 127: I think the default should instead be `html-modified-examples-only`. I don't think most contributors will think about including the example tags when they modify an example.

tools/documentation/github/deploy_doc.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if [ "${DEPLOY_TYPE}" == "stable" ]; then
+    MSG="${VERSIONTAG}"
+else
+    HEAD_COMMIT_MESSAGE=$(git log -1 --format=%s)
+    MSG="https://github.com/hidimstat/hidimstat/commit/${COMMIT_SHA} : ${HEAD_COMMIT_MESSAGE}"
+fi
+git -C ${CLONED_DOC} commit -m "${DEPLOY_TYPE} docs: ${MSG}"
+
+git -C ${CLONED_DOC} push origin main
```
GrituX, line 72: So a new PR will push the documentation on main ?

tox.ini
```diff
@@ -0,0 +1,51 @@
+;  See https://tox.wiki/en/stable/
```
GrituX, line 1: I'm not sure I completely understand what tox is used for. IIUC, it's some kind of environment manager ?

.github/workflows/build-docs.yml
Remi-Gau: ok will remove this then

.github/workflows/build-docs.yml
Remi-Gau: wait I think you don't need this: it's just because nilearn uses plotly for plotting some interactive stuff so need some browser set up

I will remove this

tools/documentation/github/build_type.py
Remi-Gau: no need to add the tag for the examples to start running: in this PR I never user the commit message to trigger an example build and the modified example where build anyway

tools/documentation/github/deploy_doc.sh
Remi-Gau: nope this gets run on push to main or on release

tox.ini
Remi-Gau: well in this case it makes it easier to make sure you build the doc the same way locally and in CI: it sets the same environment (same python and dependencies version) and runs the same command

tools/documentation/github/build_type.py
GrituX: Ok, thanks for the clarification

tox.ini
GrituX: We have a script that sets up the environment in `tools/documentation/circleci/setup_virtual_environment.sh` through `uv` and the toml configuration file. I would rather not duplicate tools and add further dependencies, if it can be done without.

Remi-Gau:
forgot to add the workflow that will trigger circleci to host the doc of the PR when the doc build is finished

adding it now

GrituX:
I didn't realize before, but another good thing is that the doc is built in parallel of tests being run. That saves a bit of time !

tox.ini
Remi-Gau: this PR actually deletes tools/documentation/circleci/setup_virtual_environment.sh

the way tox is set up, it will also use uv

GrituX: Ok !

Remi-Gau:
modified a couple of examples to reduce number of warnings

tox.ini
Remi-Gau: @GrituX let me know if you want to keep the bash script for set up instead: I don't and to impose my design choices on a repo some else will maintain long term

Remi-Gau: other theoretical advantage of tox: it works on all OS, whereas the bash scripts tend to more easily exclude windows contributors (though in this case tox is calling make which is not always an easy thing to play with on windows)

bthirion (approved):
LGTM, thx.

.github/workflows/trigger_hosting_on_pr.yml
```diff
@@ -0,0 +1,78 @@
...
+#
+# .. seealso::
+#
+#   https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+#
+# .. admonition:: CIRCLE_CI_TOKEN
+#   :class: warning
+#
+#   The ``CIRCLE_CI_TOKEN`` used in this workflow must be a personal token and not a project token:
+#   https://support.circleci.com/hc/en-us/articles/360060360811-CircleCI-API-v2-Returns-404-Not-Found
```
Remi-Gau, line 20: @GrituX 
after merging a token will need to be set on circle CI for the deploy on PR to work

.circleci/config.yml
lionelkusch: The test report is also a way to keep a trace about the performance of the library.

.github/workflows/ci_publish_call.yml
lionelkusch: The tests reports are merged in same time in the documentation.

tox.ini
GrituX: True. Then let's roll with tox.

.github/workflows/trigger_hosting_on_pr.yml
GrituX: Ok. It's a personal token, which means that one of the current maintainers has to do it. I can do it for the moment, but I think it's best if it's one from @bthirion in the long run.

tox.ini
lionelkusch: If you want to use tox, you should also improve the documentation on how to build the documentation: https://github.com/mind-inria/hidimstat/blob/main/tools/documentation_developer/building_documentation.rst

lionelkusch (changes requested):
The actual documentation about the documentation generation the CI will be outdated can you update it?

https://github.com/mind-inria/hidimstat/blob/main/tools/documentation_developer/CI_documentation.rst
https://github.com/mind-inria/hidimstat/blob/main/tools/documentation_developer/developer_guidelines.rst

.github/workflows/circleci-artifact.yml
```diff
@@ -1,54 +1,22 @@
...
-    name: Run CircleCI artifacts reports
-    steps:
-      - name: GitHub Action step
-        uses: larsoner/circleci-artifacts-redirector-action@master
-        with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-          api-token: ${{ secrets.CIRCLE_CI }}
-          artifact-path: 0/documentation/test_reports/index.html
-          circleci-jobs: get-tests-reports
-          job-title: Check the reports of the tests here!
```
Remi-Gau: let's keep this for now and focus the PR on just the doc

.github/workflows/ci_publish_call.yml
Remi-Gau: will revert this change: let's focus the PR just on the doc

.circleci/config.yml
```diff
@@ -1,146 +1,80 @@
...
-      - checkout
-      - run:
+    get-tests-reports:
+        docker:
+        - image: cimg/base:2025.05
+        environment:
+        - GITHUB_ARTIFACT_URL: << pipeline.parameters.GITHUB_RUN_URL >>
+        steps:
+        - checkout
+        - run:
```
Remi-Gau: @GrituX 
it may be a good idea to add yml formatter to the repo because otherwise adding an indent to yml adds noise to the diff

GrituX: Agreed. What do you use with nilearn ?

Remi-Gau:
> The actual documentation about the documentation generation the CI will be outdated can you update it?
> 
> https://github.com/mind-inria/hidimstat/blob/main/tools/documentation_developer/CI_documentation.rst https://github.com/mind-inria/hidimstat/blob/main/tools/documentation_developer/developer_guidelines.rst

I have updated the dev doc a bit.

.circleci/config.yml
Remi-Gau: ymlfmt via precommit

https://github.com/nilearn/nilearn/blob/7db15590a6eb3dc8c1d441ab2f13343148c3518f/.pre-commit-config.yaml#L59-L64

Issue #779, opened 2026-09-14
[DOC] The docstring of the feature importance functions is not being rendered properly


rcap107:
### Please confirm that you have searched existing issues in the repository.

- [x] Yes

### Issue with current documentation:

The documentation of all functions in https://hidimstat.github.io/stable/api.html#feature-importance-functions is not being parsed properly. 

I believe this is due to how the _aggregate_docstring function in docstring.py is re-rendering the documentation after parsing it. 

### Idea or request for content:

The docs should be fixed so that they can be rendered properly.

### Additional context (e.g. screenshots, GIFs)

<img width="637" height="615" alt="Image" src="https://github.com/user-attachments/assets/142c8e8a-8d79-49ad-82b1-3e5cf7f1c2c2" />

GrituX:
Fixed by #801

PR #801, opened 2026-09-16
[API] Depecrated method_importance functions

### LINKED ISSUE #783: Deprecate the feature importance functions?
I preface this by saying this is an outsider's perspective on the library, so of course the decision about this is up to the maintainers. 

I was looking at the code of the feature importance functions (like `cfi_importance`), and it looks like they are just wrappers around `fit_importance` and pvalue_selection: all the parameters in the init and the fit are added as possible parameters for the function. 

From the discussion during the sprint at Inria #768, my understanding is that the functions are meant to facilitate how users can access the feature importances. I would argue that the current implementation has more downsides than upsides:

- From what I can see, the functions are written manually, and they accept all the parameters that can be passed to each method. There are so many! A new user would get lost in the docstring of the function just as much as they would in the docstring of the class, and it's actually even worse because in the class fit parameters are separate. If the functions are meant to be a simpler entry point, then some of the burden on the user should be removed, possibly by setting the least important parameters (whichever they may be) to some good default.
- Functions return selected features, importances and pvalues, but not the model: this means that a user that fits the model to select the features (possibly an expensive operation) would end up with the selected features, but not with the actual model. I am not familiar with how these methods are used in practice, but I can imagine that users may want to pickle the fitted model, which is far less practical if it's stored inside a function rather than being the model by itself.  
- This is more on the maintainer's side, but the functions add a lot of maintenance overhead: each function is code that is mostly duplicated with the code of their respective class, meaning that any time that one of the parameters in the base method changes, the same parameter should also be updated in the function. Additionally, the docstring of the function is built mechanically by taking the docs of the class and its methods. The resulting docstring is not very useful because it does not really explain how the function works and how its return value is different from the class. Additionally, building the docstring requires a lot of convoluted code that also needs to be maintained #779 

I think it would be better to deprecate the functions and instead focus on the user guide to immediately show users how they can access the same information (selected features, importances, pvalues) without all the additional overhead.

GrituX:
I added a `deprecated` note to the docstring of all functions `method_importance`, so that we remove them at v0.6
I had to modify the _aggregate_docstring functions to add the `Note` sections, and modified the tests accordingly.


bthirion:
you probably want to add tests for these warnings.

jpaillard (commented):
Can we move the deprecation warning at the begining of the API doc?

bthirion (approved):
LGTM, thx !

GrituX:
> Can we move the deprecation warning at the begining of the API doc?

Unfortunately not, I tried several different methods and it just breaks the rest of the formatting even when doing modifications to the `_aggregate_docstring` function.

GrituX:
This also closes #779

GrituX:
> > Can we move the deprecation warning at the begining of the API doc?
> 
> Unfortunately not, I tried several different methods and it just breaks the rest of the formatting even when doing modifications to the `_aggregate_docstring` function.

Nevermind, I managed to do it. I think it's better this way, the info is directly visible.

jpaillard (approved):
Great thx!

